PR #84, opened 2018-11-05
[W.I.P] Server side rendering

oleast:
This adds the capability to use server-side rendering with React.
The aim is to render the 'first meaningful paint' on the server, and then let React the over after it has loaded. This can potentially increase loading performance by a lot.

What has been done:
- Create a simple express web server to render the JSX in the back-end.
- Refactor the webpack config to a more modular style.
This lets us maintain it a bit more easily, and lets us reuse parts for different main configs. This also includes a webpack config for the server.
- Reconfigured styling to compile to a CSS file instead of loading it inline to the  Javascript. This lets the page load the styling before the components. This removes the problems with un-styled components loading before the CSS has been initialized.
- Add environment configuration to differentiate Server, a client with SSR and client without SSR.
- Set up cookies to hold the selected eventView instead of localStorage. Makes it usable for the server as well.
- Make sure fetch works on both front-end and back-end (isomorphic fetch polyfill).
- Make sure the client can be rendered with and without an SSR backend. Most development should probably be done without SSR, as it is kind of a pain to set up.

- Created a decorator which and static method for fetching data on the backend before anything is rendered. Works with a request cache to make back-end render (with data) below 100 ms for frequently visited routes.

_Note: Does no longer require Redis_


oleast:
First comment has been updated with new information, should be re-read 😄

PR #111, opened 2018-11-14
ImageView transitions

R33an1:
Transitions for the ImageView events


oleast:
We are doing a small redesign of these components, and will have to think about how we can implement hover transitions after the changes.

PR #112, opened 2018-11-15
feat: Add TS type checking plugin to webpack config.

sklirg:
This makes it easier to spot type errors during development,
as they will be output by webpack during `yarn dev`.

Note: This uses tsc for compiling the *.tsx? files, not babel, which
means that there might be inconsistencies in module resolution[0].

[0]: https://github.com/Realytics/fork-ts-checker-webpack-plugin#modules-resolution


sklirg:
It runs asynchronously from the webpack build, so while the type checking might be slow, the code change happens just as fast as before.

PR #125, opened 2018-11-21
Improve readability of Nibble statistics chart

torjusti:
Attempt at improving the readability of the Nibble statistics chart.

This takes us from

image

to

image

. Some work is still needed on the overall responsiveness of the page, but the chart should scale.


src/profile/components/Statistics/Orders/orders.less
```diff
@@ -15,6 +15,7 @@
 }
 
 .centerChart {
+  height: 550px;
```
oleast: Would this not force the size of all the charts, kinda breaking the card for the pie chart?

torjusti: You're absolutely correct, didn't notice it because it only looked *slightly* broken :innocent:

torjusti:
Agreed, there you go!

PR #126, opened 2018-11-21
Redirect user to origin page after login

torjusti:
Redirect the user to the page from where the login was performed, instead of showing a mostly blank page containing *Logged in*.


src/authentication/components/AuthCallback.tsx
```diff
@@ -16,7 +17,15 @@ class AuthCallback extends Component<IProps> {
   }
 
   public render() {
-    return <div>Logged in</div>;
+    if (this.props.auth && this.props.auth.user && this.props.auth.user.state) {
```
oleast: Maybe clean this up a bit?
Like fetching `auth` from `this.props`to avoid writing `this.props` this many times?

Issue #120, opened 2018-11-20
24 Hour format on events


Paalar:
On event pages the time format is 12H without AM/PM. Since we live in Norway it would be nice to have it be a 24H format instead, or at least displaying AM/PM.

Example:  https://beta.online.ntnu.no/events/614

oleast:
Fixed by #127

PR #127, opened 2018-11-23
changing timedate format in events to 24Hr

kharann:
fulfilling issue #120 24 Hour format on events

Hopefully, this fix will work fine on any other computers. Would be great if someone could check if this is a universal fix on the problem or just a lucky coincidence due to my computer settings  or something

EDIT: i just followed https://moment.github.io/luxon/docs/manual/formatting.html#table-of-tokens for the toFormat() since it seems like its the correct manual

BEFORE:

not24hr

AFTER:

24hrfix1
24hrfix2
24hrfix3


kharann:
would you want an util only for converting ISO strings into this date format or should you include a format variable  string to specify preferred format since in AttendanceEvent.tsx they use "d MMM HH:mm " for the complete date while in picture card they separated the Date and time

PR #132, opened 2018-11-23
Remove Google Plus from footer

henrikhorluck:

- QA / Code Review

## Code Checklist

- [x] The code is ready to be merged

- [ ] I have updated the build configuration
- [ ] These changes requires changes to configuration in production 
    - [ ] I have applied the required changes in production
    - [ ] I cannot apply the required changes in production before this is deployed.

## Description of changes
Remove Google Plus from the footer, add CSS-rule to make the footer look more 'natural' according to how it was before.

This also fixes an issue if PR [#2179](https://github.com/dotkom/onlineweb4/pull/2179) in OW4 is merged, because that deletes the Google Plus icon in OW4.

## Screenshot(s) if appropriate
**BEFORE:**
image
**AFTER:**
image

### DIFF
--- a/src/core/components/Footer/footer.less
+++ b/src/core/components/Footer/footer.less
@@ -25,6 +25,10 @@
     margin-right: 20px;
     opacity: 0.7;
 
+    &:last-child {
+      margin-right: 50px;
+    }
+
     img {
       height: 32px;
     }
@@ -116,8 +120,4 @@
 #footer-map .gmnoprint,
 #footer-map div a img {
   display: none;
-}
-
-.social > a:last-child {
-  margin-right: 50px;
 }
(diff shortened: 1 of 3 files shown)


henrikhorluck:
Fixed the linting issue, all good now?

src/core/components/Footer/footer.less
```diff
@@ -117,3 +117,7 @@
 #footer-map div a img {
   display: none;
 }
+
+.social > a:last-child {
+  margin-right: 50px;
+}
```
oleast: Sorry about not getting this in the first review, but if you have a look at the nested selector at line 21, there is already code for selecting almost this element. Nesting the ``:last-child` in there would provide for a bit of a cleaner solution?

henrikhorluck: Agree, fixed it now

Issue #121, opened 2018-11-20
Single card displaying on whole page


Paalar:
On the 'Karrieremuligheter' page
- When there is a single card it displays the whole page
- Until there are four or more cards on the page the first card will extend further down on the page.

PR #136, opened 2018-11-23
fixing 1 card view in career thingies

kharann:
#121  
Fixing the issue where if there's only 1 card left after filtering it will display the whole page.

<--- BEFORE                          AFTER --->
singlecard
twocards
4 Cards looking as normal
4cards

### DIFF
--- a/src/career/less/career.less
+++ b/src/career/less/career.less
@@ -75,7 +75,7 @@
 
 .jobList {
   display: grid;
-  grid-template-rows: 1fr;
+  grid-template-rows: 1fr 1fr 1fr 1fr;
   grid-gap: 20px;
   grid-area: jobs;
 }
--- a/src/career/less/career.less
+++ b/src/career/less/career.less
@@ -75,7 +75,7 @@
 
 .jobList {
   display: grid;
-  grid-template-rows: 1fr 1fr 1fr 1fr;
+  grid-template-rows: repeat(4, min-content);
   grid-gap: 20px;
   grid-area: jobs;
 }


src/career/less/career.less
```diff
@@ -75,7 +75,7 @@
 
 .jobList {
   display: grid;
-  grid-template-rows: 1fr;
+  grid-template-rows: 1fr 1fr 1fr 1fr;
```
oleast: As you can see in the images, this will result in the first 4 elements in the list having the exact same size; the size the will fit the largest of the first 4 elements, while the rest will have the size they should have. This is because `1fr 1fr 1fr 1fr` will create 4 boxes of equal size, to fit the largest, no matter how many elements are actually present in the DOM. Example:
career-filler-grid
Note that this will also create all of that whitespace on mobile, which will just be a lot of whitespace before the footer.


To resolve this, we could hack it a bit, and force the size of the first 4 elements to fit their own min-height:
``` less
grid-template-rows: repeat(4, min-content);
```
Since the content is nothing when there are no extra elements, their height will be nothing.

Still a bit hacky, but it will show as it is supposed to. But if the filter bar is made larger, the problem will still arise again. If you find a better solution, I'm up for suggestions 😄

kharann: Seems like that way seems to work fine. I didnt get that problem when the filter bar gets bigger. It seems to look just fine. if a whitespace issue on mobile ever becomes a problem we could always just use @media. Maybe a pic of what happends when the filter bar gets bigger

oleast: Seems I was wrong about expanding the height of the filter menu, it did not result in any issues for me when testing. I think we will have to keep this thing in mind when the internship application season starts after the summer, to look out for potential problems.

PR #138, opened 2018-11-23
Fix #131 - Contact mail on events

### LINKED ISSUE #131: Contact mail on events
On events in the bottom right corner there is a 'contact' card displaying what committee is responisble for the event and how to contact them. At the moment it displays the correct committee, but the mail is written as 'komite@online.ntnu.no'. Would be nice if it was automatically displayed as 'arrkom@online.ntnu.no' if arrkom was responisble for the event. 

screenshot 2018-11-23 at 14 18 25

henrikhorluck:

- QA / Code Review

## Code Checklist

- [x] The code is ready to be merged

- [ ] I have updated the build configuration
- [ ] These changes requires changes to configuration in production 
    - [ ] I have applied the required changes in production
    - [ ] I cannot apply the required changes in production before this is deployed.

## Description of changes
Fix Issue #131: Makes the email-address for organizer dynamic, also with an anchor tag, makes the default email hovedstyret@online.ntnu.no

### DIFF
--- a/src/events/components/DetailView/Contact.tsx
+++ b/src/events/components/DetailView/Contact.tsx
@@ -11,7 +11,9 @@ const Contact = ({ event_type, organizer_name, company_event }: INewEvent) => {
   let organizer_email: string = 'hovedstyret@online.ntnu.no';
   if (organizer_name.match(regex)) {
     organizer_email = organizer_name.substring(0, 3).toLowerCase() + 'kom@online.ntnu.no';
-  }
+  } else if (organizer_name === 'seniorKom') {
+    organizer_email = organizer_name.substring(0, 6).toLowerCase() + 'kom@online.ntnu.no';
+  };
   const mailtoString: string = 'mailto:' + organizer_email;
 
   return (
--- a/src/events/components/DetailView/Contact.tsx
+++ b/src/events/components/DetailView/Contact.tsx
@@ -7,7 +7,7 @@ import style from './detail.less';
 const Contact = ({ event_type, organizer_name, company_event }: INewEvent) => {
   const color = getEventColor(event_type);
 
-  const organizer_email = `${organizer_name.toLowerCase()}@online.ntnu.no`;
+  const organizerEmail = `${organizer_name.toLowerCase()}@online.ntnu.no`;
 
   return (
     <div className={style.contact}>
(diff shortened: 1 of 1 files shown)


src/events/components/DetailView/Contact.tsx
```diff
@@ -7,14 +7,16 @@ import style from './detail.less';
 const Contact = ({ event_type, organizer_name, company_event }: INewEvent) => {
   const color = getEventColor(event_type);
 
+  const organizer_email = `${organizer_name.toLowerCase()}@online.ntnu.no`;
```
oleast: Make it camelCase, and I'll merge 😄

src/events/components/DetailView/Contact.tsx
```diff
@@ -7,14 +7,16 @@ import style from './detail.less';
...
+
   return (
     <div className={style.contact}>
       <CardHeader className={style.detailHeader} color={color}>
         Kontakt
       </CardHeader>
       <Block title="Arrangør">
         <p>{organizer_name}</p>
-        <p>komite@online.ntnu.no</p>
+        <a href={`mailto:${organizerEmail}`}>{organizer_email}</a>
```
oleast: Remember to change it here as well

src/events/components/DetailView/Contact.tsx
henrikhorluck: But that doesn't make sense when all the other variables are in snake_case?

src/events/components/DetailView/Contact.tsx
henrikhorluck: Ah, my bad

src/events/components/DetailView/Contact.tsx
oleast: While I agree that it is a bit strange, it is a result of our current technologies.

JavaScript and TypeScript use camelCase, and we use it as much as possible. But all objects from the back-end are in snake_case because that is how Django and Django Rest Framework does it by default, this is because Python uses it. It can be changed in the API, with something like [drf-camel-case](https://github.com/vbabiy/djangorestframework-camel-case), which just transforms it when requests are made.

Doing that with a current API with a lot of applications depending on it is not really viable. To do something like that, we would have to make a new API (v2), which would just be the same API with camelCase, which would be strange.

A common way of handling this is to remap the objects which are received from the back-end when they are fetched (and posted). But this just results in a lot of unnecessary extra code, _in my opinion_.

We are therefore left with somewhat of an abomination.
When the back-end is slowly replaced with other services, this will not be as much of an issue.

Issue #60, opened 2018-10-31
Fix display of event image view on smaller screens


blauks:
If you look at the image view of the events on a smaller screen the three big event viewers (that are supposed to group the different events) are shown on top, followed by the 9 other events (that are supposed to be split up into the different groups). I think the best way to present the image view is to first have one of the big event viewers, and then have the 3 corresponding smaller ones that are in the same group. And then the same for the 2 others.

mariush2:
This was done in #213, closing.

PR #139, opened 2018-11-23
creating a imageView event version for lower screen widths

kharann:
Fixing issue #60.

I mean this does solve the issue, not sure if this is a fine way of solving this issue or if its a better way. Maybe just take a look and give some feedback. This will trigger if the width is lower than 1024px

BEFORE, AFTER:
widthfix

Doesnt break the normal view:
normalwidth


oleast:
I think we will have to do another go at this after the redesign has been implemented

PR #148, opened 2018-11-24
Clean up typings for static contextType

oleast:
- Upgrade the `@types/react` version to add support for static contextType
- Replace all manual assignments of `this.context` typing with one for each class.


oleast:
Types are not matching yet. There are no types available for React versions between `16.4` and `16.7`.
ContextType has been manually declared until a future release of React, React Router and types for both are matching.

Issue #130, opened 2018-11-23
Add polyfills for outdated browsers


oleast:
Some people are not updating their browsers for some reason.

Currently known polyfills which are needed:
- `Array.prototype.flatMap`
- `AbortController`

oleast:
Resolved by #151

PR #151, opened 2018-11-28
adding polyfills #130

kharann:
Hopefully this is the correct way of adding polyfills.  Issue #130 . Please tell me if there's something i should change


oleast:
I think polyfills should be added in a way that you don't need to think about them being there.
How that would normally work is that you just import the polyfill, like:
``` typescript
import 'polyfill-package-name';
```

What this then does is to run a function attaches the polyfill itself to the global scope of the application, `window` and `self` for browsers, and `global` for Nodejs. But it only does so if the javascript engine does not yet support that feature.

A good example would be for [AbortController Polyfill](https://www.npmjs.com/package/abortcontroller-polyfill#how-to-use). Which does not attach itself if AbortController already exists on `self` or `global`  ([code](https://github.com/mo/abortcontroller-polyfill/blob/master/src/polyfill.js#L7-L9)).

Maybe we should add a `polyfills.ts` file, which can be imported by the index.ts file?

kharann:
I do think making a polyfills.ts would probably simplify the process of having to keep importing it when using them in other files. Not sure if it would have a big impact on performance or not, but if its not that big of an issue it could be a good idea. 

I just changed the way I imported the polyfills to import at the top of the file. The reason I asked is because TSLint kept screaming at me when I used require so I wondered if I was supposed to not use require at all

oleast:
Since polyfills attach to the global scope, they will only need to imported once. They should actually not be imported more than once. That is much of the reason that a single file with polyfills would be helpful.

As of right now, this branch does not build at all, since a package that is imported is not present. `'abort-controller' !== 'abortcontroller-polyfill'`. The latter one is the one we want, as the first is not a polyfill, I think.

When it comes to `flatMap`, it is included in the `core-js` package. This should do the trick.
``` typescript
import 'core-js/fn/array/flat-map';
```

oleast:
@duvholt 
Babel env for this project has set `useBuiltIns` to `'usage'`, which is supposed to add the needed polyfills based on their usage and the configured env.
However, this does not seem to add `flatMap`, since it is a `stage 3` proposal, but it has already been shipped with some browsers (current versions of Chrome and Firefox).

Babel preset env [docs](https://babeljs.io/docs/en/babel-preset-env#shippedproposals) has a section for such things but it does not seem to actually add anything from my testing?

How would we make it add `flatMap` for this project without specifying it manually?

duvholt:
Oops, I see you've already done exactly as I suggested! :snail: 

Looks like Babel docs suggests to manually import it as you said: https://babeljs.io/docs/en/v7-migration#remove-proposal-polyfills-in-babel-polyfill-https-githubcom-babel-babel-issues-8416
So basically disregard everything I said.

However, I would suggest to be more specific in your browserslist as it now includes lots of garbage browsers: 
```
❯ npx browserslist       
and_chr 70
and_ff 63
and_qq 1.2
and_uc 11.8
android 67
android 4.4.3-4.4.4
baidu 7.12
bb 10
bb 7
chrome 70
chrome 69
edge 18
edge 17
firefox 63
firefox 62
ie 11
ie 10
ie_mob 11
ie_mob 10
ios_saf 12
ios_saf 11.3-11.4
op_mini all
op_mob 46
op_mob 12.1
opera 56
opera 55
safari 12
safari 11.1
samsung 7.2
samsung 6.2
```

For example using `> 1% in NO` returns a much more sane list:
```

kharann:
> Since polyfills attach to the global scope, they will only need to imported once. They should actually not be imported more than once. That is much of the reason that a single file with polyfills would be helpful.
> 
> As of right now, this branch does not build at all, since a package that is imported is not present. `'abort-controller' !== 'abortcontroller-polyfill'`. The latter one is the one we want, as the first is not a polyfill, I think.
> 
> When it comes to `flatMap`, it is included in the `core-js` package. This should do the trick.
> 
> ```ts
> import 'core-js/fn/array/flat-map';
> ```

i changed it to core-js but i got a small issue/error?
```[ts]
Could not find a declaration file for module 'core-js/fn/array/flat-map'. 

'/home/anhkha/Documents/Coding/onlineweb-frontend/node_modules/core-js/fn/array/flat-map.js' implicitly has an 'any' type.

  Try `npm install @types/core-js` if it exists or add a new declaration (.d.ts) file containing `declare module 'core-js/fn/array/flat-map';```

i already ran `npm install @types/core-js`. But it seems like its still there

sklirg:
Regarding browserslist, I suggest reading this short article: https://jamie.build/last-2-versions

Furthermore, you can use this tool to see what you match (the same as @duvholt showed in his comments running the browserslist script): https://browserl.ist/?q=%3E0.25%25%2C+not+ie+11%2C+not+op_mini+all

PR #161, opened 2018-12-12
fixes committee mail if member is not a committee member

kharann:
When a member is not a part of the committee it will say "null@online.ntnu.no". This will just replace this with an empty string. Haven't been able to test if this works since I can't access the profile page to test it.

### DIFF
--- a/src/profile/components/Profile/MainProfile.tsx
+++ b/src/profile/components/Profile/MainProfile.tsx
@@ -11,6 +11,12 @@ import Progress from './Progress';
 export interface IProps {
   user: IFullProfileUser;
 }
+const committee_mail = (mail: string): string => {
+  if (mail) {
+    return `${mail}@online.ntnu.no`;
+  }
+  return '';
+};
 
 export const MainProfile = ({ user }: IProps) => (
   <>
@@ -20,7 +26,7 @@ export const MainProfile = ({ user }: IProps) => (
         <Content title="Kontakt">
           <KeyValue k="Telefon" v={user.phone_number} />
           <KeyValue k="E-post" v={user.email} />
-          <KeyValue k="Komité-e-post" v={`${user.online_mail}@online.ntnu.no`} />
+          <KeyValue k="Komité-e-post" v={committee_mail(user.online_mail)} />
         </Content>
       </Pane>
       <Pane>
--- a/src/profile/components/Profile/MainProfile.tsx
+++ b/src/profile/components/Profile/MainProfile.tsx
@@ -11,12 +11,8 @@ import Progress from './Progress';
 export interface IProps {
   user: IFullProfileUser;
 }
-const committee_mail = (mail: string): string => {
-  if (mail) {
-    return `${mail}@online.ntnu.no`;
-  }
-  return '';
-};
+
+const committeeMail = (mail: string) => mail ? `${mail}@online.ntnu.no`  : ''
 
 export const MainProfile = ({ user }: IProps) => (
   <>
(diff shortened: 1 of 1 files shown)


oleast:
Very nice!

For this I would think that rendering the `Komité-e-post` header would also be unnecessary?
In that case we could go with something like:
``` typescript
{user.online_mail && <KeyValue k="Komité-e-post" v={`${user.online_mail}@online.ntnu.no`} />}
```

Another thought could be to change the `KeyValue` component itself to not render if the value is `falsy`? This would need some checking on other uses of the component, but it might be OK for it to work that way?

oleast:
Another thing to think about here is the reason for this occuring in the first place.
This should have been picked up by Typescript, but it was not, because the API-model is wrong.

In this case, the `IFullProfileUser` model describes the `online_mail` as a `string`.
The actual API responds with something more like:
``` typescript
online_mail: string | null;
```

We'll have to look into this a bit more, as I suspect more of them are wrong. But it would be nice to clean this one up, could you do that @kharann ? 😄

kharann:
> Very nice!
> 
> For this I would think that rendering the `Komité-e-post` header would also be unnecessary?
> In that case we could go with something like:
> 
> ```ts
> {user.online_mail && <KeyValue k="Komité-e-post" v={`${user.online_mail}@online.ntnu.no`} />}
> ```
> 
> Another thought could be to change the `KeyValue` component itself to not render if the value is `falsy`? This would need some checking on other uses of the component, but it might be OK for it to work that way?

Im not really sure how to make the KeyValue not render when the mail is null. I've been tinkering around a little and an alternative, but quite ugly method, would be making the return value an array with the header and mail. Im quite new in react so maybe point me in the direction of some documentations or what i should be looking for

oleast:
> Im not really sure how to make the KeyValue not render when the mail is null. I've been tinkering around a little and an alternative, but quite ugly method, would be making the return value an array with the header and mail. Im quite new in react so maybe point me in the direction of some documentations or what i should be looking for

React has some good documentation for [conditional rendering](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator)

One solution could be to rewrite the value type to allow `string | null`,
then conditionally render the entire component something like this:
``` typescript
export const KeyValue = ({ k, v }: IProps) => (
  { v && <div className={style.kv}>
    <p>{k}</p>
    <h3>{v}</h3>
  </div> }
);
```

This would render the HTML if the value is 'truthy' (which all strings except `''` are), and render nothing if the value is 'falsy' (like `null`, `undefined`, `0`)

Though I think the solution in the message you quoted may be better, as it has not unintended side-effects

sardred:
Hey y'all, haven't touched the codebase in a good long time, but here are my 0.02 euro.
If the [online_mail]((https://github.com/dotkom/onlineweb-frontend/blob/6e37b1ba45ed5f43b110996b4157f3b7e05e444d/src/profile/models/User.ts#L11)) field is either a string or null, wouldn't it be better to mark it as [optional](https://www.typescriptlang.org/docs/handbook/interfaces.html)?

oleast:
> Hey y'all, haven't touched the codebase in a good long time, but here are my 0.02 euro.
> If the [online_mail](https://github.com/dotkom/onlineweb-frontend/blob/6e37b1ba45ed5f43b110996b4157f3b7e05e444d/src/profile/models/User.ts#L11) field is either a string or null, wouldn't it be better to mark it as [optional](https://www.typescriptlang.org/docs/handbook/interfaces.html)?

Nice to see you here 😄 

From how I understand it, optional is a shorthand for 'maybe `undefined`';
``` typescript
foo?: string;
// is the same as:
foo: string | undefined;
```

Should be changed to `string | null`, as that is what the API gives us.

oleast:
The solution works, and other problems can be fixed later.

kharann:
Yeah, I've been trying to find a way to do this. But it's kinda hard for me to see if my solution actually works since I'm still unable to access the profile site when I run yarn Dev. Which is why I'm quite uncertain if some of the things I wanna test out works😅

PR #163, opened 2019-01-08
Add Weekday to Event Detailview

henrikhorluck:
Listen. Weekdays are hard. We all are terrible at remembering stuff, but this will highly likely help people that don't know which weekday they are signing up for, reducing the chance of people ending up with colliding schedules, leading to the O very scare "Prikk" 🔴 😱 😱 😱 

(Do note there should maybe be added some more styling, depending on screen size and overlap)

The fancy dancy comparison picture:
BEFORE:
screenshot 2019-01-08 at 19 03 49

AFTER:
screenshot 2019-01-08 at 19 04 37


sardred:
It looks kinda cramped under "påmelding", and a bit messy under "oppmøte". Would it look better to add a newline between the weekday and the rest? In that case you'd have space to spell out the weekday entirely too.

oleast:
I think this would be very useful as well!

But I also think things like this should be created as issues first, instead of just implementing it.
We need a little discussion on how we would design this, as I think the dates are already a bit 'much', which is the reason we removed the year from the date.

One thing is to have the information you want, another is to have too much information on the screen, which makes everything hard to find.

We also have the problem of making the 'Påmelding' card too large, which in the current configuration makes the card gaps above and below the 'Oppmøte' and image card disproportionate.
This is mostly because of the rule-bundles right now, but it would happen in a lot more cases if another line is given to the weekday.

PR #179, opened 2019-01-28
[W.I.P] Create a module to show information about the Online Office

oleast:
Module/app for the Online office.

Aims to show the calendar, today's servants and maybe some form of a leaderboard for Nibble.

Uses the express backend as a proxy to wrap the google API, instead of giving all freedom with our Google keys to our users.

I think generalizing the EventCalendar would abstract all meaning away from it, to the point where it would not be worth it. Some of the code from the EventCalendar is instead copied, while other parts are abstracted and reused.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,6 +13,7 @@ import HttpError from './core/components/errors/HttpError';
 import { Route } from './core/components/Router';
 import Frontpage from './frontpage';
 import Hobbys from './hobbygroups';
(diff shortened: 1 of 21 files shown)


plusk:
What's the status on this @oleast?

oleast:
> What's the status on this @oleast?

I would like for this to be implemented in a way in the future, it would at least be nice. But I think the way I did this is flawed, and I think it should be closed.

PR #195, opened 2019-02-18
fix: Make sure all PostCSS settings are applied.

sklirg:
Configure all PostCSS settings in postcss.config.js rather than some in css.config.js and some in postcss.config.js, as they overwrite each other.

Left: prod today, right: this fix
image


sklirg:
Props to @sardred for the tip regarding autoprefixer being able to fix this, it uncovered that autoprefixer might not be correctly set up.

Issue #164, opened 2019-01-10
Shrink huge front page images


niklasmh:
I have a feeling that some images on the front page can grow too big on certain widths:
![image](https://user-images.githubusercontent.com/8504538/50991256-8962e600-1514-11e9-909b-f9a5ce0b9499.png)

It is possible to fix this setting a `max-height` and use `object-fit: cover` (or contain) on the img-tag to make the image scale in right proportions.

This results in (with `object-fit: cover`):
![image](https://user-images.githubusercontent.com/8504538/50991648-76044a80-1515-11e9-80ca-6635c2fb21a3.png)

Or `object-fit: contain`:
![image](https://user-images.githubusercontent.com/8504538/50991625-6127b700-1515-11e9-89ba-c35e5c612c2c.png)

niklasmh:
The solution on this problem is transferable to #60.

niklasmh:
@mariush2 Any status on this?

mariush2:
I am doing a re-redesign of the frontpage views and detail views. So that will include a fix to this.

mariush2:
This was confirmed as done with niklasmh.

PR #200, opened 2019-02-28
New event-views!

mariush2:
This pull-request contains a 're-styling' of all the event views we currently have. This pull-request also includes a fix to #164.

Here are some screenshots of the different changes I've made.  __Any more input as to how this re-styling should look would be greatly appreciated!__

## Image view
bilde

## List view
bilde

## Calendar view
bilde

## Detailed view
bilde


mariush2:
# New list view
bilde

henrikhorluck:
I would like if you also implemented weekdays to the detailed view, since you are tinkering with it already, as in https://github.com/dotkom/onlineweb-frontend/pull/163

src/frontpage/components/Articles/articles.less
```diff
@@ -59,3 +68,44 @@
...
+
+@media (max-width: 610px) {
+  .articles {
+    width: 425px;
+  }
+}
+
+@media (max-width: 485px) {
+  .articles {
+    width: 320px; // Because iPhone 5
```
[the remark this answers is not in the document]
henrikhorluck: It has the same screen size as the iPhone 5S and iPhone SE, the latter which is only turning 3 years old on March 31., so I would say it makes sense to support it

niklasmh:
I actually really agree with @henrikhorluck. Weekdays can even be shortend to two letters; mo, tu, we, to, fr, sa, su - which will not affect the design choices that much.

Also, it does seem logical to not repeat dates unless the event wraps two days or more. There are not many cases where we need two separate dates. Maybe we can be a little smarter with dates. I am just mentioning the idea as the current design is still affected by the old OW4 and it may not need to be that way. This probably need some more discussion and probably a Figma sketch too.

sardred:
Consider aligning the event name text in the list view to the left, might be more legible. I'd maybe like the attendance column to be aligned to the right, not sure why. Maybe because they're numbers, or because they're the last column of the "table". Not sure how it would look, because then the person-icon would be out of line on "Låvefest" for example. 
I like that the list uses the space it has available now, but I still prefer the more "block of color" approach to the previous design. That's easily settled in a quick "what do we like better" at the next meeting though, typical figma stuff.

mariush2:
# New detailed view with weekdays!
bilde

oleast:
I have a lot to say, but a lot of it is my opinions as I may be a bit (over) attached to the design.

I can see you've done a lot of great work here!
I also think we should take this chance to iron out what we actually want this to look like in the end.

Just to get a start, I have a few thing I would like to point out about the proposal:

### List view:

- Text color sticks a bit out on the page. Should change it to the `@darkGray` configured in the color profile. Should be configured for the icons as well.

- I think list view text should be aligned to the left, as Anders said.

- What happened to the status indicators/icons? I know they were not correctly implemented yet, but they are/were an essential part of the pitch for a new design and event logic.

They were to be used to show a set of statuses for the user related to the event:
- If the user wants to attend the event (would work for non-attendance event as well, so they can appear in your personal calendar and you can receive notifications).
- If your are on the waiting list
- If you are attending
- disable if there is no relation to the user?

- This does not at all work for mobile users, it starts overflowing the margin on the right at anything less than `1200px` in width.

- I'm not really ready to part with the full color event-type thingy for the list view. But I can't really make it look good, so I should maybe leave that off, but I might at least write it down.

### Detail view

- I think it looks _way_ better with the list view header on top.
It also has a quite important use-case, which is showing the event-type, title, and status of the event!
It was not implemented correctly according to the design specs, as the redundant date and amount of attendees should not be displayed there as well.

I also like the asymmetric approach bit bit more, it is useful as the 'påmelding' card can have a varying size depending on the amount of information in it.

- The gap between all cards should be the same in all directions.

### Image view:

We've had a few problems with the color of the event image clashing hard with the event-type color, which is a bit bad. One solution to that could be to detach the event-type at the top from the event card itself, and floating it above the image a bit? Should at least look into it.



All in all, well done, @mariush2 !!

henrikhorluck:
The view is broken at around 1136 pixels width:
Screenshot 2019-03-17 at 19 36 08

The list view is not correctly centered, but that might have to do with the whole page being a tad bit wider than the window. This is also an issue for the iPhone 5/SE view, where it looks like the 

There is an issue with too long titles for some events, like 
"Ekskursjon 2020: Infomøte og uformell avstemning" which goes on too many lines in the image-view
Screenshot 2019-03-17 at 19 41 45

The view-type buttons on the top should move to their more mobile-friendly size earlier, or have just a tiny bit more margin between them and the "Arrangementer" title. In the list views, while the placement of "Ekskursjon" looks great when the title spans two lines, it would look much better for one-line-titles if it was just a little lower
Screenshot 2019-03-17 at 19 42 18

mariush2:
> The view is broken at around 1136 pixels width

I can't seem to replicate this on my computer.

> There is an issue with too long titles for some events[..]

Fixed this by switching `height: 4rem` to `min-height: 4rem`

> The view-type buttons on the top should move to their more mobile-friendly size earlier

I added a small margin as you suggested.

henrikhorluck:
Correction, the issue I have with  the view being broken is at exactly 1024 pixels width, sorry for not giving the correct amount, I would think it has something to do with the text CSS and the other CSS changing being one pixel off

mariush2:
The issue with the image-view at 1024px should be fixed now.

src/events/components/ImageView/image.less
```diff
@@ -22,7 +22,7 @@
   margin-top: 20px;
 }
 
-@media (min-width: @owTabletBreakpoint) {
+@media (min-width: 1025px) {
```
sardred: You should be able to use `@owTabletBreakpoint + 1` or similar with less if that's the intended breakpoint

sardred: There are also two other breakpoints (600px and 320px) that are not attached to a variable atm :)

mariush2: Changed to `(min-width: @owTabletBreakpoint + 1)` as you suggested. The breakpoints you mentioned are specific to this page, so I don't see why they should/would be added to the variables we currently have.

mariush2:
# Mobile screenshots of current version

## Image view
imageview

## List view
lisview

## Calendar view
calendarview

## Detail view
detailedview

# Notes about this version
The iPhone 5/SE version _should work_. But there is an issue when the event-title is super long, such as on the event "Næringslivsnettverket KiD". Not sure how we would go about solving this, the best would probably be to add a breakpoint somewhere in the text if a single word in the event-title is too long.

mariush2:
#  Desktop screenshots of current version

## Image view
desktopimage

## List view
desktoplist

## Calendar view + detailed view
calendar_detailed

tor0405:
oooooouf!

mariush2:
# Fix tablet image view

The last commit fixes #164. Made the image-view shrink when it first breaks. Goes back to full-width on the `600px` breakpoint.

Screenshot 2019-03-28 at 11 04 26

oleast:
This has come very far, @mariush2.
Very nice work!

I still have one large 'irk' about the detail-view.
I don't think the header above the image should be removed at all.

First of all, I think it looks a lot better to have it there.
But there is also a problem with the information it presented not being available anymore.
the header is supposed to display:

| event-type (with name) | title | user-status |
|------------------------|-------|-------------|

In this current incarnation, the event type is only subtly hinted at through the header colors. One of the reasons for this entire project is to show them more prominently.

The header is stashed into the already large amount of text in the description.

The user status is entirly neglected in this redesign.
This is also a large reason for this projects existance, while the semantics are not quite ready, I think it is a mistake to get rid of it.

src/events/components/DetailView/detail.less
```diff
@@ -44,6 +51,10 @@
   width: 50%;
 }
 
+.blockGrid > div > p {
```
oleast: selectors for `.blockGrid` could be written in a more Less-like way?

Not really part of this pull-request, just noticed it while readying through

``` less
.blockGrid {
  display: flex;
  flex-wrap: wrap;
  ustify-content: center;

  > div {
    width: 50%;

    > p {
      text-transform: capitalize;
    }
  }

  > .fullBlock {
    width: 100%;
  }
}
```

sardred:
I agree that we shouldn't lose relevant information, but I'd prefer it presented in a different manner than what the current header does. The "copy-paste item from list view" looks a bit placeholder imo. Brainstorm on Figma!

mariush2:
# Changed the detailed view

Added the listview block to the top of the detailed view as requested. I kept the title in the body, as I think the title and the accompanying body text should be placed next to eachother.

bilde

## Regarding the status icons

In my opinion the user status should be done at a later stage seeing as the new views don't really adhere to the current implementation. Maybe work them into the detailed view somehow without using an icon or with an icon + some describing text?

I fully agree getting completely rid of the status icon is a bad idea, I just think that this particular feature is outside of the scope of this PR.

src/events/components/DetailView/PictureCard.tsx
```diff
@@ -13,6 +13,7 @@ const PictureCard = ({ image, event_start, event_end, location, company_event, e
   const startTime = DateTime.fromISO(event_start).toFormat('HH:mm');
   const endDate = DateTime.fromISO(event_end).toFormat('d MMM');
```
henrikhorluck: I would like to remind you that we should follow the official guidance on how to write dates, according to [Språkrådet](https://www.sprakradet.no/sprakhjelp/Skriveregler/Dato/)

mariush2: Good catch, added a fix to this in the latest commit.

PR #210, opened 2019-04-06
Increase z-index of dropdown-menu

henrikhorluck:
- To make it override the lightning in our loading-spinner

Fixes the issue we were informed about on by mail


mariush2:
Seems good to me. Merging.

PR #211, opened 2019-04-09
Add event-filtering to archive

### LINKED ISSUE #196: Add filtering options for events
Users should be able to search and filter events.

There should be a component at the top of the `/events` page.
It should contain options for filtering with at least the following options:
- Name
- Company
- Event Type
- Date Range
- Full text search on title, ingress, description and location?

henrikhorluck:

- QA / Code Review

Fixes #196 

## Code Checklist

- [ ] The code is ready to be merged

## Description of changes
Add Event-filtering module, as described in the issue, so far the idea is to have text search, which searches basically all text related to the event, and time range input, maybe even dropdowns to choose specific types of events. 

TODO:
- [x] Find out how to filter events
- [x] Style the stuff (#228)
- [x] Add dropdown for event types
- [x] Find out how to fetch all events

Probably smart to make a context for the search input, and use that in `src/events/providers/EventsRepo.tsx`, and return the filtered events as `eventList`

## Screenshot(s) if appropriate
Not styled/not ready


src/events/components/EventsHeader/index.tsx
```diff
@@ -13,41 +13,42 @@ export interface IProps {
...
+    case EventView.LIST: {
+      return <ListViewIcon />;
+    }
+    case EventView.IMAGE: {
+      return <ImageViewIcon />;
+    }
+  }
+};
+
+const changeViewIconButton = (viewType: number, changeView: (view: EventView) => void, view: EventView) => (
```
oleast: Functions returning JSX should be FunctionComponents.
This will allow for better optimization for React.

Simply returning JSX from a function call will make the operation happen on every render of the parent component. Making it its' own component will let us tweak that, to for example only render if the props to the component change.

src/events/components/SearchModule/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  searchText,
+  timeStart,
+  timeEnd,
+  onTextInput,
+  onTimeEndInput,
+  onTimeStartInput,
+}) => {
+  return (
+    <>
+      <DateInput label={'Fra: '} time={timeStart} onChange={onTimeStartInput} />
```
oleast:
``` typescript
label={'Fra: '} -> label="Fra: "
```

src/events/hooks/useEventsRepoState.ts
```diff
@@ -51,12 +53,40 @@ export const useEventsRepoState = () => {
     updateEventList(events);
   };
 
+  const getFilteredEventList = () => {
```
oleast: This operation will happen on every single render of the component.
That may lead to performance degradation quite easily.

We may be able to mitigate this by:
- Consider subscribing to the Context outside of the function.
- Not writing the entire function inside of the Component/function body. Write a function by itself that takes the eventList as a parameter.
- Using React.useMemo() to only run the function when the relevant variables have been changed.

I'll give a little example of how `useMemo` can be used to make things faster:
``` typescript
// An expensive operation which will slow down rendering a lot.
const func = (foo: number) => {
	return expensiveOperationThatTakes1second(foo);
};
```

Component without `useMemo`:
``` typescript
// It will take a least 1 second to re-render every single time,
// because of the expensive 'func' operation.
const Component = ({ foo, bar }) => {
	const baz = func(foo);
	return (
		<div>
			<span>{baz}</span>
			<span>{bar}</span>
		</div>
	);
};
```

Component **with** `useMemo`:
``` typescript
// This component will ONLY run the expensive operation when the 'foo' variable has changed. 
// To useMemo we give a simple function to wrap the operation, and what is called a 'dependency-array' ([foo]).
// On every render, useMemo checks if the elements of the array have changed since the last time.
// if the have, re-run the function. If they are the same, return the same result as last time without running the function.
const Component = ({ foo, bar }) => {
	const baz = useMemo(() => {
		return func(foo)
	}, [foo]);
	return (
		<div>
			<span>{baz}</span>
			<span>{bar}</span>
		</div>
	);
};
```

src/core/providers/Cookies.tsx
```diff
@@ -1,21 +1,30 @@
...
+  timeEnd: DateTime;
 }
 
 const initializeCookies = (inital: { [name: string]: any }): ICookies => ({
   ...inital,
+  frontpageEventView: getFrontpageEventView(inital.eventView),
   eventView: getEventView(inital.eventView),
+  searchText: '',
+  timeStart: DateTime.fromJSDate(new Date(new Date().setDate(new Date().getDate() - 690))),
+  timeEnd: DateTime.fromJSDate(new Date(new Date().setDate(new Date().getDate() + 690))),
```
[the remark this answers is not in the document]
henrikhorluck: Fixed in upcoming commit

src/events/components/SearchModule/index.tsx
henrikhorluck: Fixed in upcoming commit

src/events/components/EventsHeader/index.tsx
henrikhorluck: So change it to something like this?
```ts
export interface IPropsChangeViewIconButton {
  changeView: (view: EventView) => void;
  viewType: number;
  view: EventView;
}

const ChangeViewIconButton = ({ viewType, changeView, view }: IPropsChangeViewIconButton) => (
  <div
    className={classNames(style.choice, {
      [style.choiceActive]: view === viewType,
    })}
    onClick={() => changeView(viewType)}
    tabIndex={0}
  >
    <ViewIcon viewType={viewType} />
  </div>
);
```

oleast: Yeah, that looks good :)

src/core/providers/Cookies.tsx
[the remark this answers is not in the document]
oleast: The default could probably be to show ongoing events + events 1 month into the future?

src/events/components/EventsHeader/index.tsx
oleast: Could also consider moving the component to a file of its' own.
may be a bit more readable in the long run, but I'll leave that up to you.

In general I think move components should have their own file.

henrikhorluck: I agree, they should be their own files, even though they are basically just a summation of other components, since having differently named `IProps` is kinda cluttered

src/core/providers/Cookies.tsx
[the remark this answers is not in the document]
henrikhorluck: Sounds good, 
we still need to change the ListView to show all events instead of 7, fetch all events, and filter the events for it to have any effect though

kharann:
Could we have a screenshot on how it looks now? :)

henrikhorluck:
This is how it looks like atm (not pretty):
Screenshot 2019-04-10 at 18 45 35
Also the filtering is not implemented fully yet, even though some of the code is there it doesn't actually filter (yet)

src/events/components/Events.tsx
```diff
@@ -0,0 +1,70 @@
...
+        view={cookies.eventView}
+        availableViews={[EventView.LIST, EventView.CALENDAR]}
+      />
+      <SearchModule
+        searchText={cookies.searchText}
+        onTextInput={onTextInput}
+        timeStart={cookies.timeStart.toFormat('yyyy-MM-dd') || '2013-01-01'}
+        timeEnd={cookies.timeEnd.toFormat('yyyy-MM-dd') || '2020-01-01'}
+        onTimeStartInput={onTimeStartInput}
+        onTimeEndInput={onTimeEndInput}
```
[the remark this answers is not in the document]
henrikhorluck: This is now done

src/core/providers/Cookies.tsx
```diff
@@ -1,21 +1,30 @@
...
 /**
  * All available cookies should de declared as optionals here.
  * This lets us have som more control of what cookies are avaiable.
  */
 export interface ICookies {
+  frontpageEventView: EventView;
   eventView: EventView;
+  searchText: string;
+  timeStart: DateTime;
+  timeEnd: DateTime;
```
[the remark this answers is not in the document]
henrikhorluck: This has been moved to the URL, with a separate context

src/events/hooks/useEventsRepoState.ts
henrikhorluck: Now in use

src/events/components/Events.tsx
```diff
@@ -37,25 +37,41 @@ const Events = ({ location, history }: RouteComponentProps) => {
   const params = new URLSearchParams(location.search);
 
   const onTextInput = (event: ChangeEvent<HTMLInputElement>) => {
-    params.set('search', event.target.value);
+    if (event.target.value) {
+      params.set('search', event.target.value);
+    } else {
+      params.delete('search');
+    }
```
oleast: These look quite generic?

Would probably be possible to create a function or hook which abstracts this away a bit?

oleast: Maybe a `useQueryParam` hook?

```
const Foo = () => {
	const [startDate, changeStartDate] = useQueryParam('startDate');
	return (
		<input value={startDate} onChange={changeStartDate} ... />
	);
};

Comment (henrikhorluck) [reply src/events/components/Events.tsx]: So you to make a generic hook that sets and reads a query-parameter? But wouldn't that go against us using the queryParamContext to set the values?

Comment (oleast) [reply src/events/components/Events.tsx]: It may do.
The thought behind putting it in a context was to type all query parameters in a good way.
It may be a bit hard to work with though.

Using something like a hook to work with query parameters could make it very easy, but you would have to do all the dirty work of converting it from a string more locally.
Would also make it easier to have collisions where different part of the app uses the same key for a different type of value.

I have no idea if this works, but it is an example of a hook for query params.
It requires that we upgrade React Router to version 5.
``` typescript
export const useQueryParam = (key: string): [string | null, (value: string | null) => void] => {
  const { location, history } = useContext(__RouterContext);

  const params = new URLSearchParams(location.search);
  const value = params.get(key);

  const setValue = (newValue: string | null) => {
    if (newValue) {
      params.set(key, newValue);
    } else {
      params.delete(key);
    }
    history.replace(`${location.pathname}?${params}`);
  };

  return [value, setValue];
};
```

henrikhorluck: But we still have the issue that the differensiert input-elements svar their value in differensiert ways, for example `<input type=«checkbox»/>` saves he state in the form of the `checked` property, and the elements `value` is pretty much useless for our use

oleast: That is how inputs work though. It is something we have to work with (against).
It can be quite useful to create wrappers for inputs, to make them easier to use, especially things like checkboxes, radio buttons, and file inputs.

henrikhorluck: I have now refactored it to use our new `useQueryParam`-function, would be great if you could take a look into it 😃, especially the placement of the functions, and moving them away from `Events` into `SearchModule`

src/core/hooks/useQueryParamsState.tsx
```diff
@@ -0,0 +1,29 @@
+import * as H from 'history';
+import { DateTime } from 'luxon';
+import { EventTypeEnum } from '../../events/models/Event';
+
+export const defaultSearchParam = '';
+export const defaultDateStartParam = DateTime.local().toISODate();
+export const defaultDateEndParam = DateTime.local()
```
oleast: These are constants, and should probably be written CAPITALIZED

src/events/api/events.ts
```diff
@@ -25,6 +25,11 @@ export const getEvents = async (args?: IEventAPIParameters): Promise<INewEvent[]
   return data.results;
 };
 
+export const getAllEventsParamtered = async (args: IEventAPIParameters): Promise<INewEvent[]> => {
+  const data = await getAllPages<INewEvent>(API_URL, { format: 'json', page_size: 80, ...args });
+  return data;
+};
```
oleast: This could probably be combined with `getAllUserEvents` in a single `getAllEvents` function where the `user` is optional (`user?: IAuthUser`)

src/events/components/SearchModule/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+        time={DateTime.fromISO(dateEnd || defaultDateEndParam).toFormat('yyyy-MM-dd')}
+        onChange={(event) => setDateEnd(event.target.value)}
+      />
+      <SelectMultiple
+        onEventTypesInput={(event) =>
+          setEventTypes(
+            JSON.stringify(
+              [...event.target.options]
+                .filter((eventType) => eventType.selected)
+                .map((eventType) => parseInt(eventType.value, 10))
```
oleast: Larger functions like this should probably not be inline in the JSX.
Try to keep as little logic as possible in the JSX, and rather as functions in the component body.

src/events/hooks/useEventsRepoState.ts
```diff
@@ -1,21 +1,57 @@
...
 import { INewEvent } from 'events/models/Event';
+import { useQueryParamsState } from '../../core/hooks/useQueryParamsState';
+import { QueryParams } from '../../core/providers/QueryParams';
+import { isAfter, isBefore, isInDateRange } from '../utils/eventTimeUtils';
 
 export type EventMap = Map<number, INewEvent>;
 
 const INITIAL_STATE: INewEvent[] = [];
 
+const getFilteredEventList = (searchContext: ReturnType<typeof useQueryParamsState>, eventList: INewEvent[]) => {
```
oleast: This hook/function is becoming quite large.
Should probably try to split some of the logic is separate functions instead of inside the function body.

Should also try using `useMemo` and `useCallback` in apropriate places to mitiagte performance issues.
Remember that the _entire_ hook and all subscribed children of the context will be read and re-rendered when _anything_ in the state is changed.

src/core/hooks/useQueryParamsState.tsx
oleast: Constants typically use underscores `_` as a delimiter, like so:
``` typescript
const DEFAULT_VALUE = 'foo';
```

henrikhorluck: I knew something was making them unreadable 🤔

henrikhorluck:
Yes, I know I fucked up the rebase by not pulling master first :))) working on it

henrikhorluck:
Current look desktop:
image

Current look mobile: 
image

PR #213, opened 2019-04-10
Fix for issue #212

### LINKED ISSUE #212: There is an issue on mobile frontpage when too few events
There is an bug on mobile when there is too few events to showcase. We should sort the events by type on mobile.

Here is a screenshot showcasing the bug:
bilde

mariush2:
# This pull request aims to fix #212.
We now have columns for "Bedpres", "Kurs" and "Sosialt/Annet". This also means that the view on mobile is the way described in #212, meaning we now show all the events of each column after one another (List-items after Card-items).

## Here is a screenshot from before:
<p align="center">
before
</p>

## Here is a screenshot after the fix:
<p align="center">
after
</p>

### DIFF
--- a/src/events/components/ImageView/index.tsx
+++ b/src/events/components/ImageView/index.tsx
@@ -82,27 +82,37 @@ export const ImageView = ({  }: IProps) => {
 
   return (
     <>
-      <div className={style.largeEventGrid}>
-        {displayEvents.eventsLeft[0] ? (
-          <LargeEvent {...displayEvents.eventsLeft[0]} />
-        ) : (
-          <LargeEventPlaceholder event_type={2} />
-        )}
-        {displayEvents.eventsMiddle[0] ? (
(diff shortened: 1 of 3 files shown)


src/events/components/ImageView/index.tsx
```diff
@@ -82,27 +82,37 @@ export const ImageView = ({  }: IProps) => {
...
+        <div className={style.eventColumn}>
+          {displayEvents.eventsLeft[0] ? (
+            <>
+              <LargeEvent {...displayEvents.eventsLeft[0]} />
+              <SmallEventColumn events={displayEvents.eventsLeft.slice(1, 4)} />
+            </>
+          ) : (
+            <LargeEventPlaceholder event_type={2} />
+          )}
+        </div>
```
oleast: This almost exact block code is repeated 3 times, consider making it a Component?

PR #215, opened 2019-04-11
replaced repeated codeblock with a component in ImageView

kharann:
Since it was pointed out that that a codeblock was repeated 3 times in #213 i though i could try my chances on turning it into a component. Feel free to give me any feedback :)

### DIFF
--- a/src/events/components/ImageView/EventColumn.tsx
+++ b/src/events/components/ImageView/EventColumn.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+
+import { IEventColumn } from 'events/models/Event';
+
+import style from './image.less';
+import LargeEvent from './LargeEvent';
+import LargeEventPlaceholder from './LargeEventPlaceholder';
+import SmallEventColumn from './SmallEvent';
+
+const EventColumn = ({ events, event_type }: IEventColumn) => {
+  return (
+    <div className={style.eventColumn}>
+      {events[0] ? (
+        <>
+          <LargeEvent {...events[0]} />
+          <SmallEventColumn events={events.slice(1, 4)} />
+        </>
+      ) : (
+        <LargeEventPlaceholder event_type={event_type} />
+      )}
+    </div>
+  );
+};
+
+export default EventColumn;
--- a/src/events/components/ImageView/EventColumn.tsx
+++ b/src/events/components/ImageView/EventColumn.tsx
@@ -7,12 +7,12 @@ import LargeEvent from './LargeEvent';
 import LargeEventPlaceholder from './LargeEventPlaceholder';
 import SmallEventColumn from './SmallEvent';
 
-export interface IEventColumn {
+export interface IProps {
   events: INewEvent[];
   event_type: EventTypeEnum;
 }
 
-const EventColumn = ({ events, event_type }: IEventColumn) => {
+const EventColumn = ({ events, event_type }: IProps) => {
   return (
     <div className={style.eventColumn}>
       {events[0] ? (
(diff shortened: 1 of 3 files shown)


src/events/components/ImageView/EventColumn.tsx
```diff
@@ -0,0 +1,25 @@
...
+const EventColumn = ({ events, event_type }: IEventColumn) => {
+  return (
+    <div className={style.eventColumn}>
+      {events[0] ? (
+        <>
+          <LargeEvent {...events[0]} />
+          <SmallEventColumn events={events.slice(1, 4)} />
+        </>
+      ) : (
+        <LargeEventPlaceholder event_type={event_type} />
```
kharann: Also wondered if it was possible to make this child component inherit event_type from its parent component or not

src/events/models/Event.ts
```diff
@@ -197,3 +197,8 @@ export interface INewEvent {
   slug: string;
   title: string;
 }
+
+export interface IEventColumn {
```
henrikhorluck: I would say it makes more sense to have this in the actual component file, because it is not a *model* for how events look like, but rather what `IProps` the component takes in, also the `event_type` should be of the type `EventTypeEnum`, not a number

src/events/components/ImageView/index.tsx
```diff
@@ -83,36 +81,9 @@ export const ImageView = ({  }: IProps) => {
...
-              <LargeEvent {...displayEvents.eventsRight[0]} />
-              <SmallEventColumn events={displayEvents.eventsRight.slice(1, 4)} />
-            </>
-          ) : (
-            <LargeEventPlaceholder event_type={1} />
-          )}
-        </div>
+        <EventColumn events={displayEvents.eventsLeft} event_type={2} />
+        <EventColumn events={displayEvents.eventsMiddle} event_type={3} />
+        <EventColumn events={displayEvents.eventsRight} event_type={1} />
```
henrikhorluck: Also the numbers here should rather be `EventTypeEnum.BEDPRES`, `EventTypeEnum.KURS` and `EventTypeEnum.SOSIALT`

src/events/models/Event.ts
oleast: ^agree

src/events/components/ImageView/index.tsx
oleast: ^agree

src/events/components/ImageView/EventColumn.tsx
oleast: It would be possible with object `rest` and `spread`, but I don't think it makes sense here.
I think this is the way to go :smile:

kharann:
So @henrikhorluck pointed out the event_type Type should be EventTypeEnum, but in events/models/Event.ts event_type has the type of number, which is why i chose the number type here. Should we change the event_type Type in the models file for consistency sake too, keep the type as number in the new component or just leave it as it is?

oleast:
> So henrikhorluck pointed out the event_type Type should be EventTypeEnum, but in `events/models/Event.ts` event_type has the type of number, which is why i chose the number type here. So before merging should we change the event_type Type in the models file for consistency sake too, keep the type as number in the new component or just leave it as it is?

Should probably change it to EventTypeEnum in Event models

src/events/components/ImageView/EventColumn.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+import { INewEvent } from 'events/models/Event';
+import { EventTypeEnum } from '../../models/Event';
+import style from './image.less';
+import LargeEvent from './LargeEvent';
+import LargeEventPlaceholder from './LargeEventPlaceholder';
+import SmallEventColumn from './SmallEvent';
+
+export interface IEventColumn {
```
oleast: I would like to have it named as props in the interface name as well.
I think every other interface in the app describes it that way, and it is nice to keep it consistent :smile:

PR #218, opened 2019-04-13
Add endless scrolling to user search in profile

oleast:
Uses an intersection observer in a hook to determine if the bottom of the list of users in visible in the view-port.
If it is, the next page of users is fetched and added.

The intersection observer is also split out into a hook, to be reused for other things.


oleast:
Currently experiencing some problems where parts of the list is loaded several times. **Should not be merged.**

oleast:
- Refactored the Profile Search Context to hooks.
- Refactored the search fetching to use a generator function. This is to make it easier to fetch a continuous list of users, and resetting parameters when the query is changed.

_Changes are now working and are prime for review/merge_

PR #220, opened 2019-04-13
[W.I.P] Add carousel to profile medals

oleast:
I was a bit lazy, so this PR depends on another: #218 

This is kinda stuck at a long going problem with the profile page.
The with seems to not be set properly somewhere, which means elements on the profile page will easily overflow their bounds.
This becomes a huge problem when you try to use a carousel the works by hiding elements based on their overflow.


plusk:
What's the status on this @oleast?

oleast:
> What's the status on this @oleast?

I've pulled my hair out a bit on this. They war the Cards/Panes work are a bit of a mistake. Even though they are easy to use and I like working with them as they describe the layout nicely.
The problem is that the result in terrible problems with sizing and widths, which can be seen all throughout the profile pages.

This results in the carousel being near impossible to implement, as it relies on overflowing the width of the page. Since the width is fucked, the carousel is fucked.

I guess I'll vote to close this and to refactor the profile pages to a better card system.

PR #222, opened 2019-04-13
Add page to view a single article

### LINKED ISSUE #28: Add page to view an article
Articles need their own view to you can read them.

### LINKED ISSUE #274: Markdown in article ingress doesn't work
**Describe the bug**
Markdown isn't rendered in frontpage article ingress

**To Reproduce**
1. Go to frontpage
2. Scroll down to article section
3. See 

**Expected behavior**
Either not allow markdown in ingress, or render markdown in ingress preview. Since markdown works inside actual articles, we should just have the frontpage follow suit. 

**Screenshots**

**Desktop:**
 - OS: Arch Linux with i3
 - Browser: Chromium 
 - Version: Version 75.0.3770.80 (Official Build) Arch Linux (64-bit)

henrikhorluck:
Fix #28 
Fix #274 

- Not planning to implement comment section with this PR atm


src/articles/components/ArticleView.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const [article, setArticle] = useState(defaultArticle);
+
+  useEffect(() => {
+    const fetchArticle = async () => {
+      const newArticle = await getArticle(articleId);
+
+      setArticle(newArticle);
+    };
+    fetchArticle();
+  }, []);
```
oleast: Should probably add `articleId` to the dependency array to make sure the article is fetched when you go from one article to another.

src/frontpage/components/Articles/MainArticle.tsx
```diff
@@ -1,12 +1,11 @@
...
 import ResponsiveImage from 'common/components/ResponsiveImage';
-import { DOMAIN } from 'common/constants/endpoints';
 import React from 'react';
 import style from './articles.less';
 
-const MainArticle = ({ absolute_url, heading, image, ingress }: IArticle) => {
+const MainArticle = ({ heading, image, ingress, id }: IArticle) => {
   return (
-    <a href={DOMAIN + absolute_url}>
+    <a href={'articles/' + id}>
```
oleast: In this case you can import `routes` from `ArticlesRouter`, and use `routes.detail`

src/frontpage/components/Articles/SmallArticle.tsx
```diff
@@ -1,13 +1,12 @@
...
 import ResponsiveImage from 'common/components/ResponsiveImage';
-import { DOMAIN } from 'common/constants/endpoints';
 import React from 'react';
 import style from './articles.less';
 
-const SmallArticle = ({ absolute_url, heading, image, ingress_short }: IArticle) => {
+const SmallArticle = ({ heading, image, ingress_short, id }: IArticle) => {
   return (
-    <a href={DOMAIN + absolute_url}>
+    <a href={'articles/' + id}>
```
oleast: In this case you can import `routes` from `ArticlesRouter`, and use `routes.detail`

henrikhorluck:
### Current design:
image
Would like some feedback!

### TODO:
- [ ] Add related articles and tags
- [ ]  Add social media links including copy link to clipboard?
- [ ] Remove the Drop Case because of poor browser support :( (screenshot from Firefox, looked good in Chrome)

src/articles/components/ArticleView.tsx
henrikhorluck: Good catch! Fixed it 😄

src/frontpage/components/Articles/MainArticle.tsx
henrikhorluck: I added it now, using `<Link>` instead of  `<a>`, but is the only way to add the `id` to it by slicing the string?

henrikhorluck:
Btw there appears to have been/is support for Vimeo videos in articles, this should probably be implementet as well

henrikhorluck:
Now with related articles:
image

henrikhorluck:
With media-queries on mobile:
image

mariush2:
Really great work! I would like it if the 'Fotograf' text is smaller and on-top of the image. Maybe with a slightly transparent background?

Haven't looked at the code yet, so can't comment on that.

henrikhorluck:
I would like it to *not* be on top of the image, but as a compromise we could move it to be along with the rest of the publication/author information?

mariush2:
To clarify I meant that we can keep the bar design we have now. But make it position absolute so it overlaps the image with a transparancy on the bar itself

henrikhorluck:
Okay turns out it doesn't look too bad, but couldn't definitively be more readable:
image

henrikhorluck:
0.6 opacity, instead of 0.5:
image

mariush2:
Try having the bar span the whole width of the image maybe?

henrikhorluck:
Width `width: 100%` (for some reason it goes beyond?)
image

henrikhorluck:
I fixed the grid so that the related articles now start with the header, what should we use this new, amazing free-space for, leave it empty, or we could place author-information etc. there
image
As a todo, it would probably be a good idea to limit the amount of related articles to be max the length of the article

oleast:
> Another thing would be to add the slug to the link, maybe so that the slug is automatically inserted after it finds the id? @oleast this is applicable to the `Event/DetailView` as well, what do you think? If people go around copying the link it would be nice to have a little more context on what is at the other end, but would perhaps be preferred if that slug _needs_ to match the id, though, which would be a bit more work.

It is nice to have context, yes. As long as it is used in the way you propose, I think it would be nice.

But I would _not_ like for it to be required at all. Going to the URL with only the id should work, and I think it makes things a lot cleaner.

henrikhorluck:
Regarding the author info, I would have preferred if the author was linked to an online-user, so that we could get a contact-the-author-email, profile-image, a short slug describing the author etc, maybe be able to link authors as commitees? But that would require backend work :(, as of right now the author is literally just a text field

src/articles/components/ArticleView/ArticleImage.tsx
```diff
@@ -0,0 +1,19 @@
...
+import IImage from '../../../common/models/Image';
+import style from './articleView.less';
+
+export interface IProps {
+  image: IImage;
+}
+
+export const ArticleImage: FC<IProps> = ({ image }) => (
+  <section className={style.articleimage}>
+    <img alt={image.description} src={DOMAIN + image.original} />
```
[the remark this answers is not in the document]
henrikhorluck: I mean if there is one place we should use the original to bask in all it's glory it would be here, and the CSS is already responsive to support basically any size

oleast:
> Regarding the author info, I would have preferred if the author was linked to an online-user, so that we could get a contact-the-author-email, profile-image, a short slug describing the author etc, maybe be able to link authors as commitees? But that would require backend work :(, as of right now the author is literally just a text field.
> 
> To be fair, it would probably not be too much work manually adding Online-users to the articles, since there is only 90

My bad though that was how it already was.

I think we should try to implement it, but it may have been implemented that way to allow _any_ author to write an article, even without a user.
May be a hassle working with and updating the field to a user, but I think we could do it :smile: 

Could help a little with bring life to articles, and a little incentive for people to write one?

Maybe it should allow multiple authors for an article though.

src/articles/components/ArticleView/RelatedArticles/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      });
+  };
+
+  useEffect(() => {
+    const fetchArticles = async () => {
+      let incomingRelatedArticleList: IArticle[] = [];
+      for (const tag of mainArticle.tags) {
+        const tagArticles = await getArticles({ tags: tag });
+        incomingRelatedArticleList = incomingRelatedArticleList.concat(tagArticles);
+      }
```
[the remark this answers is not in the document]
henrikhorluck: I have implementet your `Promise.all()`, but we cannot integrate it with the OW4 API since if you send a list of tags there you will get only the articles with **all** the tags on it: [Example](https://online.ntnu.no/api/v1/articles/?tags=%C3%B8konomi&tags=fondstyret)

src/articles/components/ArticleView/ArticleImage.tsx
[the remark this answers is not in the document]
oleast: Yeah, I agree that we should utilize the images to their full potential, but then we end up with monstrosities like [this boi here at 8.4 MB](https://online.ntnu.no/media/images/responsive/344acda7-05f2-4d30-94ba-4ffa5a019177.png). The `lg` version is at 950 KB.

Serving 8.4 MB at initial load is not optimal at all, and will break mobile users. Links to the article should not load that large image either.

We can use the `original` image, but it should be loaded after the initial load, or at the very least only on screen sizes which have use for them via `srcSet`, as described in the wiki.

src/articles/components/ArticleView/RelatedArticles/index.tsx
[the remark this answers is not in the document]
oleast: I think we could at the very least consider modifying the filter, or adding a new one for filtering on multiple tags. Like a `tags__in` or something?

Not really a need for it currently though, this work fine.

src/articles/components/ArticleView/RelatedArticles/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+      .sort((a, b) => Date.parse(b.published_date) - Date.parse(a.published_date))
+      .filter((article) => article.id !== mainArticle.id);
+  };
+
+  useEffect(() => {
+    const fetchArticles = async () => {
+      const relatedArticlePromises = mainArticle.tags.map(async (tag) => getArticles({ tags: tag }));
+      const newRelatedArticles = await Promise.all(relatedArticlePromises);
+      const updatedRelatedArticleList = updateRelatedArticleList(
+        newRelatedArticles.reduce((prev, next) => prev.concat(next), [])
```
oleast: [Array.prototype.flat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) is your friend

src/articles/components/ArticlesRouter.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Switch } from 'react-router-dom';
+import { ArticleView } from './ArticleView';
+
+const BASE_URL = '/articles';
+
+export const routes = {
+  detail: BASE_URL + '/',
+};
+
+const EventsRouter = ({}) => (
```
oleast: > EventsRouter

src/frontpage/components/Articles/SmallArticle.tsx
```diff
@@ -1,21 +1,23 @@
+import React from 'react';
+
 import { IArticle } from 'articles/models/Article';
 import classnames from 'classnames';
```
oleast: Classnames is an imported package

src/articles/components/ArticleView/RelatedArticles/index.tsx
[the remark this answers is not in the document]
henrikhorluck: I think the current solution is fine, but I added an issue for the filter we want: https://github.com/dotkom/onlineweb4/issues/2238, so that someone can add it at a later point

src/articles/components/ArticlesRouter.tsx
henrikhorluck: hehe

sardred:
ye ye ye its ya boy YUNG JUDGEMENT

### Meta

Haven't looked at the code, so no comment on that this time. I agree with Big Ole (Anders) Stokker Senpai - hereafter dubbed referred to as B.O.S.S. -  on doing design stuff should be figured out on Figma first, especially back-and-forths can be worked out pretty quick during a work session. These PRs are getting hella big as it is, 38 emails for a single PR is a lot. I'm sure I'll be repeating a lot of the stuff B.O.S.S. already said, but here goes.

### Panes

Part of a larger discussion, but I'm personally leaning away more and more from the whole paper card look as we go, it's starting to feel a bit dated. Maybe it's just the gray color we use for the background being a bit dark, but it sure is an easy way to visually separate content. Maybe a somewhat hybrid approach like [Aftenposten](https://www.aftenposten.no/) of having papers and grids within papers could fix some of my grievances. It could be cool to have a special case for articles of pumping up the margin to give that "text on a piece of paper" feeling, but i dunno, Figma that shit. 

### Legibility

Quick side-note on the comfort of reading, look into what an optimal paragraph width for them eyes we got is. I never looked too deep into it, but I'm sure there are [tons of variables](https://ux.stackexchange.com/a/3623) involved.

### Information

Move the most relevant information (author and published time) to the top, right under the image in this case. Having an "about the author" spot could be cool, typically I think I've seen that underneath the actual article, give it a cute little image with a big ol' name and some flavor text. That text could be set by the author per article for more comedic potential, or highlighting more relevant information about them. Or we can keep it simple for the users and use some existing bio field, a special field for authors, imagination is the limit. 

Following the "about the author", I'd prefer seeing related articles (and potentially other articles by that author). If I'm reading an article, I'm not interested in other ones until I'm done with it. If I do decide halfway through that I want to go to a different one, I wouldn't want something related to the one I lost interest in, so then I'd just go back to the main page. **Point is**: I prefer related articles below the article. If we have more space in width than we know what to do with (likely), then I wouldn't weep over seeing them on the side.

### Text

Consider the options available for styling, I think we use font weight in a decent way in the live version. Italics is worth considering for ingress, or smaller blocks like "about the author", which is less related to the article and then more suited to be hella flair. 

On a larger scale note again, it would be cool to introduce a serif font to the website. It adds a lot of visual separation, some styyyyyyle, and it evokes the classic metaphores of print articles etc.

### Color

I agree that grayscale can seem a bit boring, but if I've clicked into an article, I'm there for the article. If added flair can enhance the experience (see: the more fancy image-heavy NRK articles of late) then that's cool, but that typically requires more custom work on the author per article. Other than that, more visual flair is typically distracting or reducing legibility.

### Images

Like B.O.S.S. said, lg is plenty big for images. Example from the live version, the images are only gonna be 624px wide at max. So then we'd want an image larger than medium, which is 405px, to avoid visual degradation. Larger images would just be downscaled anyways, so our large images (720px) is more than enough. Clicking on the image to open up the original in a new tab would be neat though.

### Final note

No harm in brainstorming by looking at inspiration. Hit up your favorite newspapers and tech blogs and take notes! Are they playing with font weight? Italics? How are they separating paragraphs? What makes the headings eye-catching? Does the article pop from the background? Does it need to? What information are they showing about the author, and where?

All in all good work, the project is coming along nicely, and I'm looking forward to looking closer at it when I get back.

henrikhorluck:
Picture of carousel (with I think the longest title we have atm):
image
With only one related article:
image
Does not appear with 0 related articles

henrikhorluck:
Current design:
Mobile:
localhost_8080_articles_9(iPhone 6_7_8)

Tablet:
localhost_8080_articles_9(iPad)

Desktop:

localhost_8080_articles_9 (1)

I think we could make the image full-width on desktop, but otherwise quite happy with where we are atm, what do you guys think?

oleast:
Great progress, it looks good!

I think we should strive to make _all_ our layouts fill the entire with of header on all sizes. I looks strange to not fill it for certain parts of our application. It would help us in creating a more unifying look/style for our 'product'.

This is what the page looks like on a regularly scaled 1080p, which is _the_ most used screen resolution in Norway:
cf8a5073-d26d-49f7-87a6-3c4e7a9c17c5
IMO. it makes it feel cramped, almost claustrophobic.

I know it is hard to find the content to make that work, but I think it could be useful to put the metadata for the article on the left side as a separate card. Could also put more data related to the writer when we get that in place?

I think you are right about taking the image out to the border on larger screen sizes. It makes it feel more in line with our other designs.

With tags, dates and such I am wondering what would be the best of highlighting the 'description' (e.g. 'Tags') or the 'content' (e.g. 'online, ntnu, informatikk'). I would think that highlighting the content is better, as the description can often be inferred from the content. Highlighting the description makes the user focus on that rather than the content.

For the tags: How would it look the make use of boxes similar to the rule bundle descriptors used on events? They should probably be made in to generic reusable components anyways.

For dates: I think it would be good for us have a few presets for date formats, such that they are presented in the same way throughout the application. Long dates like the ones displayed here should be the same as for Events for example. This does not need to be fixed here, as there are probably other disparities, but it is just a thought :smile:

henrikhorluck:
I disagree about needing to fill the entire width, but you are free to make something on Figma to convince me otherwise, but do note, I do not want the image to be as wide as our current implementation, because then you only see the image at first when you go to the article.

oleast:
> I disagree about needing to fill the entire width, but you are free to make something on Figma to convince me otherwise.

I think it is a more fundamental thing of making our design consistent.

> I do not want the image to be as wide as our current implementation, because then you only see the image at first when you go to the article.

I agree. The current width of the article itself is probably good.

plusk:
I agree with moving author info out to the side to make the main card less cramped, and to fill out our "page width". After that, I think the PR should be merged, and then instead look at it again later down the line. We have many other "placeholder" designs, but I think having the article view in the beta so people can get more of a feel for it is important.

henrikhorluck:
Hopefully final design changes:
<details>
<summary>Desktop:</summary>
localhost_8080_articles_4
</details>

<details>
<summary>Tablet:</summary>
localhost_8080_articles_4(iPad)
</details>

<details>
<summary>Mobile:</summary>
localhost_8080_articles_4(iPhone 6_7_8)
</details>

mariush2:
Good work. I think this is ready (without having looked at the code). There should be border-radius on the image for the desktop version aswell though.

PR #228, opened 2019-04-16
[W.I.P] Styling for the event-filtering page

mariush2:
# [WIP] This PR styles the event-filtering page
I have not yet done the mobile-version of this page. Im thinking that will be done at a later stage.

Here is a screenshot of how it looks now:
search


oleast:
I'll have a look at it once it has been rebased :smile:

mariush2:
# Rebased
Screenshot 2019-04-17 at 11 26 41

# Regarding the date-inputs
I would love it if we could use [this](https://react-day-picker.js.org/examples/selected-range). It is alot more user-friendly than having two date inputs when really what we want is a range of dates.

# Regarding the layout
I feel what I have in this comment is better than the previous, however I still have trouble finding out where the dropdown menu fits best. Tried having it on the top row, but it quickly became super cramped. We should implement the new date-selector before we decide where to place stuff.

Also the EventHeader is _not_ centered.

yarn.lock
```diff
@@ -1311,7 +1358,7 @@
   resolved "https://registry.yarnpkg.com/@types/range-parser/-/range-parser-1.2.3.tgz#7ee330ba7caafb98090bece86a5ee44115904c2c"
   integrity sha512-ewFXqrQHlFsgc09MK5jP5iR7vumV/BYayNC6PgJO2LPe8vrnNFyjQjSppfEngITi0qvfKtzFvgKymGheFM9UOA==
 
-"@types/react-dom@^16.0.11":
+"@types/react-dom@*", "@types/react-dom@^16.0.11":
```
oleast: This can cause problems with type mismatches, watch out.

mariush2:
# Rebased again
This is the current state of the styling. Still a bit unsure of the placement of the different elements. Also there is a bug where the event-types filter is not working as expected.

## Screenshot of current version
styling

oleast:
I don't think we should add `moment` to this project. We already have a library for working with dates, and moment is _large_

mariush2:
> I don't think we should add `moment` to this project. We already have a library for working with dates, and moment is _large_

Removed `moment` and wrote my own formater to show the dates the same way.

src/events/components/SearchModule/DateRangeInput.tsx
```diff
@@ -82,7 +82,7 @@ export const DateRangeInput: FC<IProps> = ({
   };
   
   const formatDate = (day) => {
-    return day.getDate() + ". " +  MONTHS[day.getMonth()] + " " + day.getFullYear();
+    return day.getDate() + ". " +  MONTHS[day.getMonth()].toLowerCase() + " " + day.getFullYear();
```
oleast: Use DateTime#toFormat ?
https://github.com/moment/luxon/blob/master/docs/formatting.md

src/events/components/SearchModule/SelectMultiple.tsx
```diff
@@ -1,25 +1,59 @@
...
+  return eventTypeNumberList.map((eventType) => ({ value: eventType, label: getEventType(eventType) }));
+};
 
-  return eventTypeNumberList.map((eventType) => (
-    <option key={eventType} value={eventType}>
-      {getEventType(eventType)}
-    </option>
-  ));
+const optionStyles = {
+  control: (styles: any) => ({ ...styles, backgroundColor: 'white' }),
```
oleast: `styles` (and other input variables) probably have a type. They should _not_ be `any`

oleast:
There are also a lot of compile time error in the code here. These should probably be picked up by the editor?

src/events/components/SearchModule/DateRangeInput.tsx
niklasmh: Just updating this comment as `moment` now is removed. How to use the `date-fns` formatter: https://date-fns.org/v1.28.0/docs/format.

```javascript
import { format } from 'date-fns';
import * as locale from 'date-fns/locale/nb';
```

```javascript
return format('D. MMM YYYY', { locale });
```

henrikhorluck: Isn't that redundant because of `DateTime.format()` inclusion in DateTime? Just use `DateTime.fromJSDate(day: Date)` on the incoming date

niklasmh: Nvm, thought that we used `date-fns` and not `luxon`. My bad. Removed my comment. (Was also a little confusing that we removed `moment` and used `luxon` instead, meaning we still use moment - just a more [modernized version](https://github.com/moment/luxon/blob/master/docs/why.md) of it.)

src/events/components/SearchModule/search.less
```diff
@@ -21,12 +95,13 @@
...
-      width: 550px;
+      width: 525px;
+      .owCard();
     }
     /* stylelint-enable */
   }
 }
 /* stylelint-disable-next-line selector-type-case, selector-type-no-unknown */
-.inputFromToTo :global DayPickerInput-Overlay {
+.inputFromToTo :global daypickerinput-overlay {
```
henrikhorluck: Why is this changed? As far as I can remember (and I put this here) the `DayPickerInputOverlay`-class is something we can't rename (it is a part of the ReactDatePicker CSS, which we cannot set ourselves), so you are just making the rule useless, and then it should be removed altogether

src/events/components/SearchModule/SelectMultiple.tsx
henrikhorluck: Yup, it's `CSSProperties`

src/events/components/SearchModule/search.less
mariush2: Might have been changed by `yarn lint:fix` ? 

I can't remember changing myself:P

src/core/hooks/useQueryParamsState.tsx
```diff
@@ -17,15 +17,14 @@ export const useQueryParamsState = (location: H.Location<H.LocationState>) => {
   const dateStart: DateTime = DateTime.fromISO(params.get('dateStart') || DEFAULT_DATE_START_PARAM);
   const dateEnd: DateTime = DateTime.fromISO(params.get('dateEnd') || DEFAULT_DATE_END_PARAM);
   const eventTypes: EventTypeEnum[] = JSON.parse(params.get('eventTypes') || DEFAULT_EVENT_TYPES_PARAM);
-  const attendanceEventsChecked: boolean = !(
-    params.get('attendanceEvents') === 'true' || params.get('attendanceEvents') === null
-  );
+  const nonAttendanceEventsChecked: boolean =
```
oleast: Would say the naming here is a bit convoluted?

henrikhorluck: It is the best description of what it does, could be worse

PR #229, opened 2019-04-18
Use React-Date-Picker instead of <input type="date">

henrikhorluck:
- Add necessary functions and refactoring
- Make it possible to import normal CSS-documents, so that the CSS for the component works
- It now works in Safari!

Current look (styling coming with #228): 
image


src/events/components/SearchModule/DateRangeInput.tsx
```diff
@@ -0,0 +1,65 @@
+import { DateTime } from 'luxon';
+import React, { FC, MutableRefObject, useRef } from 'react';
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import 'react-day-picker/lib/style.css';
```
henrikhorluck: I hope there is a better way to add this CSS, since I had to add CSS-importing to webpack-config for this to work

src/events/components/SearchModule/dateRangeInput.css
```diff
@@ -0,0 +1,21 @@
...
+.InputFromTo .DayPicker-Day--end {
+    border-top-right-radius: 50% !important;
+    border-bottom-right-radius: 50% !important;
+}
+.InputFromTo .DayPickerInput-Overlay {
+    width: 550px;
+}
+.InputFromTo-to .DayPickerInput-Overlay {
+    margin-left: -198px;
+}
```
henrikhorluck: These styles could maybe be moved to a Less-document, but only `.InputFromTo` is defined by `DateRangeInput`, and as such the classes don't have the hashes at the end, which are added when using less, so we would need to find a way for the other classes to not have added hashes

src/events/components/SearchModule/DateRangeInput.tsx
kharann: Are you maybe able to import that css into a css file with
`@import "/node-modules/react-datepicker/lib/react-datepicker.css"`and load it?

henrikhorluck: The problem is that webpack wasn't able to load CSS files with import, until I added it with this PR, if I import it through a `.less`-file (which I think you are meaning) it would add the hashes to the CSS-classes, and thus the CSS would have no effect

mariush2:
The only thing I can see that is missing before we can merge this is localization. [Here](http://react-day-picker.js.org/examples/localization) is the documentation from react-day-picker. 

We should also decide if we want one clickable range input or two like we have now. [Here](http://react-day-picker.js.org/examples/localization) is an example of what I mean by one range input component.

oleast:
I have a few thoughts, but I'll have time to write it down just yet

src/events/components/SearchModule/DateRangeInput.tsx
oleast: This is an _ok_ way to import external CSS, I guess.
Doing to through Less is cumbersome because it is configured with CSS Modules.

Just make sure it works with SSR.

src/events/components/SearchModule/DateRangeInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+  handleToDateClick,
+  handleFromDateClick,
+  handleResetClick,
+}) => {
+  const modifiers = { start: dateStart.toJSDate(), end: dateEnd.toJSDate() };
+  const toInput = useRef(null) as MutableRefObject<null | DayPickerInput>;
+
+  return (
+    <div className="InputFromTo">
+      <button onClick={handleResetClick}>Rest</button>
```
oleast: Rest?

src/events/components/SearchModule/DateRangeInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <DayPickerInput
+        value={dateStart.toJSDate()}
+        placeholder="Fra"
+        format={'YYYY-M-D'}
+        dayPickerProps={{
```
oleast: `dayPickerProps` makes up a large portion of the body here.

It could be moved outside of the JSX body at least?

something like this?:
``` typescript
const fromMonthProps: DayPickerProps = {
    selectedDays: [dateStart, { from: dateStart, to: dateEnd }],
    disabledDays: { after: dateEnd },
    toMonth: dateEnd,
    modifiers: { start: dateStart, end: dateEnd },
    numberOfMonths: 2,
    onDayClick: onDayClick,
};
```

src/events/components/SearchModule/dateRangeInput.css
oleast: [You might be looking for :global](https://github.com/css-modules/css-modules#usage-with-preprocessors)

oleast: Would _very_ much like for this to be a Less file, not CSS.
It would be a bit more readable it it is nested IMHO

src/events/components/SearchModule/search.less
```diff
@@ -1,4 +1,4 @@
-.dateInput {
+/* .dateInput {
```
oleast: Why still have this large piece of code if it is not used?

src/events/components/SearchModule/index.tsx
```diff
@@ -28,41 +29,63 @@ const SearchModule: FC = () => {
...
-      </label>
-    </div>
+        <SelectMultiple
+          onEventTypesInput={onEventTypesInput}
+          eventTypes={JSON.parse(eventTypes || DEFAULT_EVENT_TYPES_PARAM)}
+        />
+        <label>
+          Vis arrangementer med påmelding
+          <ToggleSwitch
+            onChange={() =>
```
oleast: Please abstain from large functions inside JSX :smile:

src/events/components/SearchModule/index.tsx
```diff
@@ -28,41 +29,63 @@ const SearchModule: FC = () => {
...
-        <ToggleSwitch
-          onChange={() =>
-            setAttendanceEventsChecked(
-              attendanceEventsChecked === 'true' || attendanceEventsChecked === null ? 'false' : 'true'
-            )
-          }
-          checked={attendanceEventsChecked === 'true' || attendanceEventsChecked === null}
+    <>
+      <div className={style.grid}>
+        <input
```
oleast: This could probably be wrapped as a general component?

henrikhorluck:
> We should also decide if we want one clickable range input or two like we have now. [Here](http://react-day-picker.js.org/examples/localization) is an example of what I mean by one range input component.

If you by two mean [this](http://react-day-picker.js.org/examples/selected-range), I would like to say that I tried that first, but the problem is that it isn't hidden by default, and takes a huge amount of space, which is why I prefer it this way, of course we could do the hiding ourselves with css, but I prefer it this way, also we wouldn't be able to do the greying of the invalid dates in that way, but of course our implementation of what a click does can change

src/events/components/SearchModule/index.tsx
```diff
@@ -28,41 +29,63 @@ const SearchModule: FC = () => {
       )
     );
 
+  const handleToDateClick = (day: Date) => {
+    const clonedDate: Date = new Date(day.getTime());
+    const datetime = DateTime.fromMillis(clonedDate.setHours(0, 0, 0, 0));
```
oleast: This all seems a bit cumbersome, is there no other way to obtain this result with just `DateTime`?

also, camelCase

src/events/components/SearchModule/index.tsx
```diff
@@ -28,41 +29,63 @@ const SearchModule: FC = () => {
...
+    const dateStartDateTime = DateTime.fromISO(dateStart || DEFAULT_DATE_START_PARAM);
```
oleast: `dateStartDateTime` -> `dateTimeStart`?

src/events/components/SearchModule/index.tsx
```diff
@@ -28,41 +29,63 @@ const SearchModule: FC = () => {
...
+
+    if (datetime > dateStartDateTime || datetime.toISODate() === dateStartDateTime.toISODate()) {
+      setDateEnd(datetime.toISODate());
+    }
+  };
+
+  const handleFromDateClick = (day: Date) => {
+    const clonedDate: Date = new Date(day.getTime());
+    const datetime = DateTime.fromMillis(clonedDate.setHours(0, 0, 0, 0));
+    const dateEndDateTime = DateTime.fromISO(dateEnd || DEFAULT_DATE_END_PARAM);
```
oleast: Same as above

src/events/components/SearchModule/DateRangeInput.tsx
henrikhorluck: Update: does not appear to be working with SSR, getting multiple errors, from rejected promises trying to fetch CSS, but the main page loads without it, `localhost:8080/events` gets a rejected promise, document is not defined when trying to check if the browser is old IE, etc.

henrikhorluck: Is it possible to import it into Less, with `:global`?

oleast: Not really. `:global` has to be applied to each class. There are postcss plugins for doing that for you, but I think that becomes too much of a convoluted solution.

src/events/components/SearchModule/DateRangeInput.tsx
henrikhorluck: Reset*

src/events/components/SearchModule/search.less
henrikhorluck: Me forgetting to remove it, and not wanting to delete the file, since it will still be used for the styling

src/events/components/SearchModule/index.tsx
henrikhorluck: I changed the input of the function to DateTime now, but I still have to use `const dateTime = day.set({ hour: 0 });`, since the JS-Dates are automatically set to 12 O'clock of the chosen day, while it should be 00:00:00, as that is how our context has them

src/events/components/SearchModule/DateRangeInput.tsx
```diff
@@ -0,0 +1,109 @@
...
+  'Oktober',
+  'November',
+  'Desember',
+];
+
+const WEEKDAYS_LONG = ['Søndag', 'Mandag', 'Tirsdag', 'Onsdag', 'Torsdag', 'Fredag', 'Lørdag'];
+
+const FIRST_DAY_OF_WEEK = 1;
+
+const LABELS = { nextMonth: 'Neste måned', previousMonth: 'Forrige måned' };
```
henrikhorluck: Just as a FYI, I added this manually, so that we don't have to add `Moment.js` as a dependency, because it is unnecessary imo

mariush2: +1 this, looked at the documentation I linked and fully agree with you

src/events/components/SearchModule/DateRangeInput.tsx
```diff
@@ -0,0 +1,109 @@
...
+    months: MONTHS,
+    weekdaysLong: WEEKDAYS_LONG,
+    weekdaysShort: WEEKDAYS_SHORT,
+    firstDayOfWeek: FIRST_DAY_OF_WEEK,
+    labels: LABELS,
+  };
+
+  return (
+    <div className={style.inputFromTo}>
+      <button onClick={handleResetClick}>Reset</button>
```
mariush2: The reset button does not seem to work when I've tested it. However I'm not sure if we really need a reset button?

henrikhorluck: I think  it would be nice to have, but not necessary, although I am more worried about why it isn't working, because even if I make it set the parameters to `null`, it only removes dateEnd, not dateStart

henrikhorluck:
Note: I will merge this so that we can style it, but we still need to fix the CSS importorganisasjonen before meeting into master

src/events/components/SearchModule/index.tsx
henrikhorluck: I think that is outside the scope of this PR

PR #234, opened 2019-04-21
Clean up types from the olden days

oleast:
As this project reaches 1 year of age, there are some parts of it that have not aged well. Especially the parts where I experimented with TypeScript models for the first time, since I did not know the language.

This PR hopes to fix some of my stupidity.


src/events/models/Attendee.ts
```diff
@@ -1,10 +1,8 @@
-import { IUser } from 'core/models/User';
 import { IAttendanceEvent } from './Event';
 import { IExtra } from './Extras';
 
 export interface IAttendee {
   event: IAttendanceEvent;
-  user: IUser;
```
henrikhorluck: Does this mean that we no longer have an Online-user linked to an attendee? How isn't that going to affect our opportunity to implement the "Show attending"-feature? (For the events with public attendee-lists)

src/events/models/Event.ts
```diff
@@ -111,46 +109,10 @@ export enum EventView {
...
-  image: string; // StaticContent URL?
-  event_type: EventType;
-  organizer: IGroup;
-
-  // feedback: Feedback
-}
-
-export interface ICompanyEvent extends IEvent {
+export interface ICompanyEvent {
   company: ICompany;
```
henrikhorluck: This is probably way outside the scope of this PR, but why is the CompanyEvent even it's own type, and not just a link to the company directly? Also wouldn't it make more sense for it to be called organisation, so that you could add collaboration with other organisations, without it being a company event? (Again way outside the scope of this PR)

src/profile/models/Penalty.ts
```diff
@@ -48,17 +48,17 @@ export interface IMark {
   mark: {
     title: string;
     added_date: IsoDate;
-    given_by: IUser;
+    given_by: null; // Requires changes in OW4 API.
```
henrikhorluck: Remember to open an issue/PR to fix this in OW4

src/events/models/Event.ts
oleast: My guess is that it comes from how `ManyToMany` relations work.
There is a table of events can be related to many companies, and the table of companies can be related to many events. To make this work, we create a table of 'CompanyEvents', which links them together.

For simple usage a ManyToManyField would work, but we have custom logic in the `save` method of the model: [example](https://github.com/dotkom/onlineweb4/blob/develop/apps/events/models.py#L854-L860)

src/events/models/Attendee.ts
oleast: Sort of a temporary fix this one. I can set up a correct model for it.

The problem with public attendee lists resides with an another issue entirely, which I will leave to the developer implementing the event registration.

This model can not be used to display those users, as it contains more information than what we want to be displayed through the API for attending users.

PR #239, opened 2019-04-23
Remove CalendarView from /Events

henrikhorluck:


src/events/components/EventsHeader/index.tsx
```diff
@@ -9,26 +9,24 @@ import style from './eventsHeader.less';
 
 export interface IProps {
   toggleAccessible: () => void;
-  changeView: (view: EventView) => void;
+  changeView?: (view: EventView) => void;
   accessible: boolean;
-  view: EventView;
-  availableViews: EventView[];
+  view?: EventView;
+  availableViews?: EventView[];
```
henrikhorluck: This should probably be removed, will do after I come home

PR #244, opened 2019-04-25
Responsive Responsive Images

oleast:
We have a lot of options for the images we display, but it gets kinda arbitrary which image is actually displayed since people have different screens width different pixel ratios.

This tries to solve the problem of choosing an image size by calculating the optimal image to display based on the current real pixel size of the image in the DOM.

### DIFF
--- a/src/common/components/ResponsiveImage/index.tsx
+++ b/src/common/components/ResponsiveImage/index.tsx
@@ -1,16 +1,88 @@
-import React, { ImgHTMLAttributes } from 'react';
+import React, { FC, ImgHTMLAttributes, useEffect, useState } from 'react';
+
+import { useBoundingRect } from 'common/hooks/useBoundingRect';
+import { getKeys } from 'common/utils/tsHacks';
+
 import { DOMAIN } from '../../constants/endpoints';
-import IResponsiveImage, { IResponsiveImageSizes } from '../../models/ResponsiveImage';
+import IResponsiveImage, {
+  IMAGE_DIMENTIONS,
+  IResponsiveImageSizes,
+  ResponsiveImageTypes,
+} from '../../models/ResponsiveImage';
+
+export type ImageType = ResponsiveImageTypes;
+export type ImageSizes = IResponsiveImageSizes;
+export type ImageSize = keyof ImageSizes;
 
 export interface IProps extends ImgHTMLAttributes<HTMLImageElement> {
   image: IResponsiveImage;
-  size: keyof IResponsiveImageSizes;
+  size: ImageSize;
+  type: ImageType;
 }
 
-export const ResponsiveImage = ({ image, size, alt, ...props }: IProps) => {
-  const defaultImage = image[size];
+/**
+ * @summary Get the key types of images sorted by width.
+ * @example ['xs', 'sm', 'md', 'lg']
+ * @param {ImageType} type
+ */
+const getSortedDimentionKeys = (type: ImageType) => {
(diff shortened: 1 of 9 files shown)


src/common/components/ResponsiveImage/index.tsx
```diff
@@ -1,16 +1,88 @@
...
+      return aWidth - bWidth;
+    })
+    .filter((t) => t !== 'wide'); // Filter wide because it is unstable, and just mimicks 'lg'.
+};
+
+/**
+ * @summary Get the most fitting image version to display on screen based on the client screen/pixel size.
+ * @param {ResponsiveImageTypes} type
+ * @param {number} displayWidth The actual width of the displayed image in real pixel in the screen.
+ * @param {ImageSize} currentSize The default/currently displayed size. Resolution should ne be downgraded.
```
henrikhorluck: Typo?

src/common/components/ResponsiveImage/index.tsx
```diff
@@ -1,16 +1,88 @@
...
+import { useBoundingRect } from 'common/hooks/useBoundingRect';
+import { getKeys } from 'common/utils/tsHacks';
+
 import { DOMAIN } from '../../constants/endpoints';
-import IResponsiveImage, { IResponsiveImageSizes } from '../../models/ResponsiveImage';
+import IResponsiveImage, {
+  IMAGE_DIMENTIONS,
+  IResponsiveImageSizes,
+  ResponsiveImageTypes,
+} from '../../models/ResponsiveImage';
```
henrikhorluck: https://github.com/dotkom/onlineweb-frontend/wiki/(Opinionated)-Structureing-and-ordering-of-imports

oleast: > Any file not residing in the same module as the current file should be imported with a root relative route.
>
> It may be a good thing to import as root relative if the path looks like this as well: ../../../models/Something; But I won't strictly condone that.

src/common/components/ResponsiveImage/index.tsx
oleast: You are correct. I probably need better linting on my spelling 🤷‍♂️

src/common/components/ResponsiveImage/index.tsx
```diff
@@ -1,16 +1,87 @@
...
+      const [bWidth] = IMAGE_DIMENSIONS[type][b];
+      return aWidth - bWidth;
+    })
+    .filter((t) => t !== 'wide'); // Filter wide because it is unstable, and just mimicks 'lg'.
+};
+
+/**
+ * @summary Get the most fitting image version to display on screen based on the client screen/pixel size.
+ * @param {ResponsiveImageTypes} type
+ * @param {number} displayWidth The actual width of the displayed image in real pixel in the screen.
```
henrikhorluck: ~pixel~ -> _pixels_ ?

src/common/components/ResponsiveImage/index.tsx
```diff
@@ -1,16 +1,87 @@
...
+ * @param {ImageSize} currentSize The default/currently displayed size. Resolution should never be downgraded.
+ */
+const getOptimalImageSize = (type: ImageType, displayWidth: number, currentSize: ImageSize): ImageSize => {
+  const dimensions = IMAGE_DIMENSIONS[type];
+  const keys = getSortedDimensionKeys(type);
+
+  /** Flag if the compared sizes are larger than the currently displayes size */
```
henrikhorluck: ~displayes~ -> _displayed_

Issue #226, opened 2019-04-16
Refactor/Rethink how we display resources


oleast:
#### The larger problem:

We need a way to display resources/links in a good way.

We have a lot of different applications, which have their own links, and can be hard to know about without being explicitly told to go there.

These are pages like:
- Committee applications (AMaCS/Supreme Guacamole): http://opptak.online.ntnu.no
- Genfors Voting (Super Duper Fiesta): https://genfors.online.ntnu.no
- Kvitteringsskjema (Receipt Form): https://kvittering.online.ntnu.no
- Splash (Glowing Fortnight): https://online.ntnu.no/splash
- Event Registration (Regme): https://online.ntnu.no/rfid

##### Other
- Techtalks: https://techtalks.no
- Smashrank: https://smash.online.ntnu.no
- Band: https://online.ntnu.no/band
- Gaem?: https://gaem.online.ntnu.no
- Jubileum: https://jubileum.online

The resources page currently resides in a sort of broken state.

The information is outdated, the design is not easy to work with and also looks a bit dated.

We should have a way to display these things, and make them available to our users.

PR #248, opened 2019-05-11
Fix images, sorting and filtering for hobbies and resources

### LINKED ISSUE #141: The hobbygroups are missing images
On the site 
https://beta.online.ntnu.no/hobbygroups each group is missing an image.
screenshot 2018-11-24 at 13 06 11

oleast:
Fixes #141


mariush2:
Good work!

## Image CSS
Currently the images look like this:
bilde

This should be fixed before merging. (Remember to round the corners of the images themselves aswell!)

oleast:
I have to say I have no idea how to make this work right now.

The images are a fixed (square) size, which works for what they are, but we need something that makes this look good at any size with any text.

There is also the read more link, which is more content that has not been displayed yet.

mariush2:
Could try detaching the images and aligning them with the text-box.

Maybe something like this?
Screenshot 2019-05-11 at 22 33 45

Just a thought. I think detaching the images is a good idea anyways.

mariush2:
# Restyled
Seperated into two cards. Would love some input on how this looks, I'm not too sure about myself.

## Desktop
desktophobby

## Mobile
mobilehobby

src/common/models/ResponsiveImage.ts
```diff
@@ -71,6 +71,24 @@ export const IMAGE_DIMENSIONS: ResponsiveImageDimensions = {
...
+  },
+  hobby: {
+    thumb: DEFAULT_THUMB_SIZE,
+    original: [Infinity, Infinity],
+    wide: [710, 710],
+    lg: [710, 710],
+    md: [710, 710],
+    sm: [540, 540],
+    xs: [360, 360],
+  },
```
henrikhorluck: It is sort of unnecessary to have to identical image-sizes, with different names. I would prefer if we just use `resource` for both

henrikhorluck: It is sort of unnecessary to have to identical image-sizes, with different names. I would prefer if we just use `resource` for both

src/hobbygroups/less/hobbygroups.less
```diff
@@ -9,26 +10,44 @@
...
+    grid-template-columns: 1fr;
+
+    & > div {
+      margin-top: 5px;
+      width: 40rem;
+    }
+  }
+}
+
+@media only screen and (max-width: @owMobileBreakpoint) {
```
henrikhorluck: Okay so I really want for there to be a dedicated tablet mode, because it changes to the mobile-view way too quickly at the moment. Like, if I open the devtools, I still want it to be in desktop-mode

src/common/models/ResponsiveImage.ts
oleast: Yeah, thought about that, but decided against it.

They are different things, and that they use the same sizes was just a bit of a shortcut when hobby groups were implemented.

oleast:
Anyone have any ideas for what to do with resources?

oleast:
Ok, I have a little proposal:
_We merge this PR without doing anything more with resources._

Doing anything with resources is kinda blocked by: #226 , which should not have to block implementing a change for hobbies.

This PR was intended to change hobbies and resources to use ResponsiveImage, such that they are not broken and with any image at all, which is does as described.

plusk:
If the images are fixed on `/resources`, then making resources a separate page up for redesign is fine by me.

PR #261, opened 2019-05-23
Add react-helmet-async for open graph-attributes

henrikhorluck:


src/core/index.tsx
```diff
@@ -1,14 +1,19 @@
...
+  <HelmetProvider>
+    <Helmet>
+      <title>Linjeforeningen Online</title>
+    </Helmet>
+
     <Header />
     <main>{children}</main>
     <Footer />
-  </>
+  </HelmetProvider>
```
oleast: Adding the Provider again here will probably make Helmet not work in SSR?

When using a context the closest provider above in the tree will be used, which would be this provider in all cases.

If a provider needs a different configuration for client and server it should be added to the `src/server/index.tsx` file and the `src/index.tsx` file.

src/server/index.tsx
```diff
@@ -82,64 +83,81 @@ app.get('/serviceworker.js', (_, res) => {
...
-          <App />
-        </ContextWrapper>
-      </Prefetched>
-    </Cookies>
-  </StaticRouter>
-);
+const initJSX = (location: string, prefetcher: PrefetchState, cookies: { [name: string]: string }): IInitJSX => {
+  const helmetContext = {};
+  return {
+    jsx: (
```
oleast: As with how the other configured tools do it, the context should probably be made outside of this function and be passed as an argument.
The function makes a bit strange by returning two values in this way, as it is just supposed to return JSX.

src/core/index.tsx
henrikhorluck: So I refactored it, by moving the `<HelmetProvider>` to `<ContextWrapper>` (which i think makes the most sense), and adding the `HelmetContext` as an optional input, and that works both with SSR and without locally for me

Issue #232, opened 2019-04-21
[Discussion] Rethinking card design and colors


oleast:
## The problem

I've had a thing bugging me about some design choices over the last few months. When making pages with the current design model, I've often hit problems with coloring pages, when to add accent colors, and how to give the page 'life'.

The main problem seems to stem from the background colors of the page.
It is way harder to combine colors with gray (#f2f2f2) than white.

I was not planning on attacking this problem, but a comment for a PR brought me to the point of at least writing this. [Comment by @sardred which sparked this](https://github.com/dotkom/onlineweb-frontend/pull/222#issuecomment-485175622)

### Why is this a large problem

This limits us to using Cards for almost every piece of content on the page, and limits our use of colors. Pages can look very much like they are in grayscale, and therefore more boring and lifeless.

It also makes it harder to do things like full page content. You get into on of two situations:
- You put the image or content directly on the gray (#f2f2f2), which can make it look strange, at least if there are many colors. ([not a good example, but it kinda shows it](https://beta.online.ntnu.no/webshop))
- You add the full-page content into a single Card. This leaves more freedom for the content, but also restricts size to withing the margins of the card, and just looks off. (No example, it has not been done as it looks bad).

Having full-page content can be useful for things like login-, error-, or info-pages. And I would very much like for it to be easier for us to do. 

## How do we fix it

I have two current thoughts about how we can do something about it, but noe are very well formed yet.

### Boring one.

Just lightening the background color could help, I have not really tested it. May not solve all the problems, but may solve some?

### The one about making new design choices

I've had a few thoughts, but again, nothing well formed. [There are some thoughts in the comment that sparked this](https://github.com/dotkom/onlineweb-frontend/pull/222#issuecomment-485175622). Figma would be a good way to start this, and trying to work with with a multitude of the pages we have could prove to be useful.

The first thing that came to my mind was to make our background white (#f2f2f2) again. It takes a tiny bit of inspiration from [Kvitteringsskjema](https://kvittering.online.ntnu.no).

Further it works by using the current Cards more as borders, rather than papers, and also removing the shadow. It would also make it easier to add sections of different colors, like the ones in Kvitteringsskjema.

![Screenshot 2019-04-21 18:09:01](https://user-images.githubusercontent.com/14319313/56472634-2660e380-6461-11e9-8da2-b39d6815604f.png)

[...]

PR #262, opened 2019-05-24
Add basic toast message support

### LINKED ISSUE #258: Add toast messages as feedback to the user
Feedback from the app or from the server does currently not display in a good way for the user.

The messages should be displayed to the user.
This can easily be done by using something like 'toast'-messages.

Toast messages are small messages that appear on the page to notify users that something has happened.
They usually appear at the edge of the screen and will last for a set number of seconds before they disappear.

Colors or accent colors can be used to display more context. Like red for warnings and green for success.

**Additional context**
An example of usage could be event attendance. Where the result of you trying to attend the event is displayed.

Another would be paying for an event or adding money to Nibble saldo.

oleast:
Adds toast messages to with a few usable settings to make it easy to display messages to the user when something is happening.

resolves #258 

_Disclaimer: The stylish button on the top are not included in this PR_
new-toast

### DIFF
--- a/src/common/hooks/useCountDown.ts
+++ b/src/common/hooks/useCountDown.ts
@@ -3,7 +3,7 @@ import { useEffect } from 'react';
 
 import { useDecrement } from './useDecrement';
 
-const DEFULAT_TICK = 1000;
+const DEFAULT_TICK = 1000;
 
 const calculateOffset = (time: DateTime, tick: number): number => {
   const endTime = time.toMillis();
@@ -17,7 +17,7 @@ const calculateOffset = (time: DateTime, tick: number): number => {
  * @param tick How often the timer should update. Defaults to every second.
  * @returns Seconds remaining until the countdown is finished.
  */
-export const useCountDown = (endTime: DateTime, tick = DEFULAT_TICK): number => {
+export const useCountDown = (endTime: DateTime, tick = DEFAULT_TICK): number => {
   const duration = calculateOffset(endTime, tick);
   const [remaining, decrement] = useDecrement(duration);
   useEffect(() => {
--- a/src/common/hooks/useCountDown.ts
+++ b/src/common/hooks/useCountDown.ts
@@ -3,22 +3,25 @@ import { useEffect } from 'react';
 
 import { useDecrement } from './useDecrement';
 
-const calculateOffset = (time: DateTime): number => {
+const DEFULAT_TICK = 1000;
+
+const calculateOffset = (time: DateTime, tick: number): number => {
   const endTime = time.toMillis();
   const now = DateTime.local().toMillis();
-  return (endTime - now) / 1000;
+  return (endTime - now) / tick;
 };
 
 /**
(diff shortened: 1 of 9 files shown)


oleast:
Compared to the gif above, the countdown now ticks in 200 ms intervals instead of a second.

src/common/hooks/useCountDown.ts
```diff
@@ -3,22 +3,25 @@ import { useEffect } from 'react';
 
 import { useDecrement } from './useDecrement';
 
-const calculateOffset = (time: DateTime): number => {
+const DEFULAT_TICK = 1000;
```
mariush2: Spelling error, shold be "default"

src/core/utils/toast/toast.less
```diff
@@ -0,0 +1,54 @@
+@import '~common/less/mixins.less';
+@import '~common/less/constants.less';
+@import '~common/less/colors.less';
+
+.toastContainer {
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  box-sizing: border-box;
+  border: 8px solid transparent;
```
mariush2: The toast-messages should have a border-radius of 3px (same as .owCard())

src/core/utils/toast/useToast.ts
```diff
@@ -0,0 +1,7 @@
+import { useContext } from 'react';
+
+import { IToastContextState, ToastContext } from './ToastContext';
+
+export const useToast = (): IToastContextState => {
```
mariush2: Burde denne være 'export default'?

oleast: I generally prefer not using default exports.
This enforces the name of the exported function and generally makes things easier to import automatically with with the editor.

oleast:
> Great work! The only thing I'm abit unsure the colours on the messages though. I think it would make more sense to have the card be white with black text. This would also make the coloured text look better aswell.

The problem with having white cards is that they blend completly in with the rest of the content on the page. They overlap in a strange way, making it seem like the page is broken.

This is a general problem that can be added to the list in #232 

I don't really know how to solve this well.

src/core/utils/toast/toast.less
oleast: This border is not visible. It is used as used as the margin around the entire list of messages.

The reason for not using a margin is that the content would overflow inside the container as it is set to a width of `100vw`, which a margin would not uphold.

oleast:
I worked a bit on making the toasts a little more in-tune with the current design.
**The GIF in the first post has been updated to reflect this**

They now use the `owCard` mixin, and uses the `darkGray` background.
Messages are now separated by a little bap between them.
The text and progress bar has also been made bold.

Colors have been turned off for now, as they are not immediately needed, and most messages work without them.

The `useToast` hook has been updated with settings. These will let you:
- Set the hook to remove the previous message created _by the same hook_ before adding a new message. (as displayed above).
- Setting the default duration and type of messages created by that hooks `addMessage` function.

Issue #206, opened 2019-03-08
Integrate with Stripe for payments


oleast:
We use Stripe payments for event attendance, nibble account and webshop (external).
We should be able to use these features from OWF.

~This requires API-endpoints that also integrate with stripe for the back-end in OW4~
API-endpoints for Transactions (Nibble) and Relations (Events) have been implemented in OW4.

oleast:
Payments have been implemented for Transactions (Adding money to use for Nibble), ad shown here: https://beta.online.ntnu.no/payments

Something similar has to be implemented for events you are attending.
I think this should be implemented when #39 has been completed.

oleast:
Fixed by #288 and #294

PR #264, opened 2019-05-28
Add support for saldo payments with Stripe

### LINKED ISSUE #253: Let users add money to their account/saldo
Users should be able to add money to their saldo.

This should be done using [Stripe Elements](https://stripe.com/payments/elements).
The best way to do this would be using [React Stripe Elements](https://github.com/stripe/react-stripe-elements)

Depends on the API from https://github.com/dotkom/onlineweb4/pull/2266

oleast:
This PR creates a _Payments_ module to handle all payment related things.
It also sets up general support for usage of Stripe.

Depends on #263 and https://github.com/dotkom/onlineweb4/pull/2272

The module displays:
- A simple form for adding money to saldo with Stripe integration.
- All transactions for adding money to user saldo.
- All purchases in Nibble ordered by purchases date.

Furthers progress on #206
Resolves #253  

*Note: Requires a new deployment of OW4 with some small fixes to payments API code*

saldo-payment

### DIFF
--- a/config/environment.config.js
+++ b/config/environment.config.js
@@ -12,6 +12,10 @@ const BASE_CONFIG = {
   OWF_SSR: 'false',
   OWF_VAPID_PUBLIC_KEY: '',
   OWF_WEBPUSH_SERVER_URL: '',
+  STRIPE_PUBLIC_KEY_ARRKOM: null,
+  STRIPE_PUBLIC_KEY_FAGKOM: null,
+  STRIPE_PUBLIC_KEY_PROKOM: null,
+  STRIPE_PUBLIC_KEY_TRIKOM: null,
 }
 
 module.exports = {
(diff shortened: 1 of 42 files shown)


henrikhorluck:
I really want this to have some tests added, because, after all we are dealing with money here.

oleast:
While I agree that testing is nice and all, testing this would be 10 literal pains in the ass to handle.

I can see a few large problems with testing this:
- We are relying on an asynchronously loaded script from an external platform, which we don't really know how will behave.
- There is also the problem of the Stripe field handling and loading unpredictably in an `iframe`.
- Then there is the whole problem of handling the 3D Secure overlays for verification.  

While testing functionality like this is nice, we are not actually dealing with money here. That part in handled in full by the back-end, and the front-end is just a guide for using that API (which itself is tested quite well if I could say so myself).

PR #266, opened 2019-05-30
Implement support for Google/Apple/Microsoft-pay

oleast:
The Stripe `PaymentRequestButton` implements a generic interface for using payment services like Google/Apple/Microsoft-pay through the browser.

This is implemented as a button that will show if the device has a configured way to use it. 

~What is left:~
- ~Reformat the fields to display properly with a button that appears conditionally.~
- ~Add a certificate for Apple Pay, test in Apple devices.~
- ~Actual testing. It is quite hard to test, as you need to work with a secure https connection with a valid (not self-signed) cert.~

*Update*:

- Some refactoring has been done to make the code a little more approachable.

- Guide for running all required services with HTTPS for local development has been created [here](https://github.com/dotkom/onlineweb-frontend/wiki/Developing-with-HTTPS)

- User interface has been updated to accommodate the extra button:

Mobile like sizes:
create-transacton-mobileish-3

Desktop like sizes:
create-transaction-desktopish-2


oleast:
*Note:* To enable Apple Pay we have to add the domain as a trusted Apple Pay domain in the Stripe Dashboard _after_ deployment to production/beta. Stripe/Apple needs to be able to read the certificate from the URL when that URL is enabled.

PR #269, opened 2019-06-04
Animate toast messages

oleast:
- Uses React Spring to animate toast messages into and out of view.
- Animate toast progress smoothly with basic CSS transitions.

*Note that the animation runs smoothly in real life, but gifs can capture the smoothness*
animated-toast

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,6 +132,7 @@
     "react-helmet-async": "^1.0.2",
     "react-loadable": "^5.5.0",
     "react-select": "^2.4.2",
+    "react-spring": "^8.0.20",
     "react-stripe-elements": "^3.0.0"
   },
   "resolutions": {
--- a/src/common/hooks/useCountDown.ts
+++ b/src/common/hooks/useCountDown.ts
@@ -1,5 +1,5 @@
 import { DateTime } from 'luxon';
-import { useEffect } from 'react';
+import { useEffect, useMemo } from 'react';
 
 import { useDecrement } from './useDecrement';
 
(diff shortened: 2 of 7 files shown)


henrikhorluck:
Okay so this is not a comment on this specific PR, but wouldn't it be a good idea to pull the Toast-messages out of this project, and add is as a package, so It could easily be reused for regme, SDF, receipt-form, fuzzy-train etc.?

oleast:
I've thought about it, and that would probably be very useful.
It would need a lot of changes to decouple the implementation from this project though.

It is a general problem that we have, as reusing things like this would be nice, but maintaining 30 different packages it a lot of work.
Could go for something like a monorepo I guess, but that would require _a lot_ of restructuring

henrikhorluck:
Would this work https://github.com/dotkom/react-components ?

mariush2:
Good stuff! I would prefer if the animation in, is the same as the animation out. As in, both do the same translation. This makes the animations more uniform.

Otherwise, really good work!

oleast:
I'm a bit unsure of what would be the best.
I've tweaked en transitions to transform to and from the same position now, but I also added a transition to the height to make the toasts not jump.

The result is this:
toast-fast-smooth

This is the result without overwrite:
multi-smooth-toast

henrikhorluck:
*Minor question:* the loading-bar has a margin against the border on the left side, but not the right, is that intentional?

oleast:
> _Minor question:_ the loading-bar has a margin against the border on the left side, but not the right, is that intentional?

It appears you've found a little iceberg. This was cause by some bad CSS on my part, and fixing it resulted in revealing that the CSS was broken to the core because of the fixed height of the animated message.

Animating boxes based on non-fixed sizes seems to be near impossible, which makes us kinda stuck. I've made it work, but its not perfect.

PR #270, opened 2019-06-12
Add customisable data-tables

oleast:
Create a common data-table which makes handling larger amounts of data a bit easier.
Lets the developer define ways to sort the table columns, and specify the headers.

Converts transactions and orders in the payments view to data-tables for exmaples:
data-table

### DIFF
--- a/src/common/components/Table/DataTable.tsx
+++ b/src/common/components/Table/DataTable.tsx
@@ -1,0 +1,99 @@
+import React, { ReactNode, useCallback, useEffect, useMemo, useState } from 'react';
+
+import { getKeys } from 'common/utils/tsHacks';
+
+import style from './table.less';
+
+export type SortFunction<T> = (a: T, b: T) => number;
+
+// tslint:disable-next-line interface-over-type-literal
+export type DataTableHeaders = { [key: string]: string };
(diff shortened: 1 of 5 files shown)


henrikhorluck:
Would love to see a triangle or something to indicate which way the column is sorted, just for more visual clarity, and perhaps a smaller min-height for the rows? I think it looks great with the symbols you  use, but if we only have text it might be a bit much.  Otherwise: 🔥 great work!

oleast:
> Would love to see a triangle or something to indicate which way the column is sorted, just for more visual clarity, and perhaps a smaller min-height for the rows? I think it looks great with the symbols you use, but if we only have text it might be a bit much. Otherwise:  great work!

I've thought about making small triangles, but I have a hard time figuring out how to place them correctly. Ideally I would like the triangle to be placed a pixel above the underline if it points up, and a pixel below if it points down. I would also like for them to be very small :man_shrugging: 

This is just problematic, and I thought I could rather add them later if we wanted.

For the min-height I've experimented a lot, and I think spacing is important. Even without the icons I think that is as small as they should go.

oleast:
Added said arrows:

sorting-arrows

PR #271, opened 2019-06-12
Design new cards for hobbies

oleast:
*Note that this merges into another in-progress feature-branch, as this does not handle what to do with resources*

I took some design inspiration from the newest version of Offline, and made new cards for hobbygroups.

There are of course ways to improve, but that would need some thinking.
- We could add colors the the card, either random at run-time or defined on the model.
- The images could float half way on the edge of the cards, or something, but I don't know how to do that.

Desktop design:
b41499fc-87b1-44ae-8946-2be70d0f26fd

Tablet design:
50fed52e-4b25-4530-b901-d6d241ae3799

Mobile design:
dd1cceba-fd9a-4aaf-96cf-6c0aa48bfbe6


src/common/less/constants.less
```diff
@@ -1,7 +1,7 @@
 /* ********** Variables ********** */
 @owDesktopMaxWidth: 1200px;
 @owTabletBreakpoint: 1024px;
-@owMobileBreakpoint: 768px;
+@owMobileBreakpoint: 767px;
```
henrikhorluck: Have you checked that this doesn't leave a 1 px fucked-up css state for our other pages?

src/hobbygroups/components/HobbyCard/Divider/index.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+import style from './divider.less';
+
+export const Divider = () => <div className={style.divider} />;
```
henrikhorluck: Does this really need it's own file? I think it is so small that it is unnecessary, but 🤷‍♂

src/hobbygroups/components/HobbyCard/Title.tsx
```diff
@@ -0,0 +1,5 @@
+import React, { FC } from 'react';
+
+import style from './card.less';
+
+export const Title: FC = ({ children }) => <h2 className={style.title}>{children}</h2>;
```
henrikhorluck: Again, does this need it's own file? Especially when you have the same `less`-document

plusk:
I don't mind images being circular, it's a design choice and we can re-make logos to be circular. I agree with not being into the whole "lines everywhere" thing, less is more etc. No offence :heart:

anywho i made a branch with a pr going into this branch with a pr going into another branch, so please have a look at #272 for my proposal, threw something together real quick instead of studying for my machine learning exam

oleast:
I think we should try to work with the circular images, and bend the hobbygroup logos to our will :smile: 

I agree on removing the line below the link, but I want to keep the lines for the content. At the very least the line below the title.
This would also make it nicer if we introduce more information about the groups.
I would like to add stuff like the group leader, members, and other info to the groups like how to contact or join

src/hobbygroups/components/HobbyCard/Divider/index.tsx
oleast: I is a larger discussion, but it is a way of writing components more like atoms (see Atomic Design).

The divider is a small and reusable component that can be used anywhere, so why should we have to make it again everywhere it should be used?

plusk:
I re-added the title divider in the PR before I merged it.

> This would also make it nicer if we introduce more information about the groups.
I would like to add stuff like the group leader, members, and other info to the groups like how to contact or join

Then we'll add that to make it nicer if/when that stuff is added. I'm not too hot on adding much more information about the hobby groups there when we're already linking to the wiki pages which should be easy enough to edit and maintain. If we were to add something in these cards then I would have it as a "show more" expanding thing, but in a grid like that I think it would look hella wack with the rows below shuffling down etc. 

If I could add a feature request: I'd like it if clicking the image would bring up a higher resolution version in a new tab.

On a somewhat related note, I'm not a huge fan of how the grid doesn't fit the page width like we've discussed in other PRs. Not sure if it's better to stretch the gaps between them.

oleast:
> I re-added the title divider in the PR before I merged it.

:hear_no_evil: :heart_eyes: 

> > This would also make it nicer if we introduce more information about the groups.
> > I would like to add stuff like the group leader, members, and other info to the groups like how to contact or join
> 
> Then we'll add that to make it nicer if/when that stuff is added. I'm not too hot on adding much more information about the hobby groups there when we're already linking to the wiki pages which should be easy enough to edit and maintain. If we were to add something in these cards then I would have it as a "show more" expanding thing, but in a grid like that I think it would look hella wack with the rows below shuffling down etc.

It was mostly just dreaming of a larger system where people could manage their own groups, being able to join and all such things. It will probably never be implemented anyways.

> If I could add a feature request: I'd like it if clicking the image would bring up a higher resolution version in a new tab.

Like when right clicking and selecting 'open in new tab', just with it using the 'original' version of the image?
I don't know how we could do that without removing the use case of clicking the images as a link for example.
i don't know know if it is possible to add things to the context menu?

> On a somewhat related note, I'm not a huge fan of how the grid doesn't fit the page width like we've discussed in other PRs. Not sure if it's better to stretch the gaps between them.

Yeah, I struggled a lot with that. Went back and forth between grid and flexbox to find the optimal way of doing this. In most cases the gutters got extremely large, or the cards were stretched out too much.
I'm guessing the optimal solution is to use flexbox and define a max and min width for the cards to stretch between, always filling the page with a fitting number of cards or whatever (without media queries :heart: ?). But I'm not currently skilled enough to get it working perfectly.

src/hobbygroups/components/HobbyCard/Title.tsx
oleast: answered in another comment

plusk:
> Like when right clicking and selecting 'open in new tab', just with it using the 'original' version of the image?
> I don't know how we could do that without removing the use case of clicking the images as a link for example.
> i don't know know if it is possible to add things to the context menu?

Not thinking context menu, more just clicking the image. Not saying it has to be baked into every image on the site or anything (though that would be dope), but in these cases where there is no other reasonable interaction with the image, I want to be able to see the original by clicking it. Typically if an image is larger, you could have a smaller "download" or similar icon in a corner of an image, maybe hidden until image is hovered. 

If we do bake that into images in general, shouldn't be too hard to override that behavior if we want to just have a straight up image link or whatever. 

> In most cases the gutters got extremely large, or the cards were stretched out too much.

I did something about this whole thing when working on the employee list for Bekk last summer I think, but that might have been just not having a static "page" width and instead adapting that too. That could be an interesting idea, probably hard to get the footer to align too, but at least the page content could be synchronized potentially. No idea, don't really have time to look at it, but I suppose we'll both have access to the employee list come Monday, which is probably way improved from when I worked on it a year ago. Should be okay to open up the inspect tool and see how it's done.

oleast:
> Not thinking context menu, more just clicking the image. Not saying it has to be baked into every image on the site or anything (though that would be dope), but in these cases where there is no other reasonable interaction with the image, I want to be able to see the original by clicking it. Typically if an image is larger, you could have a smaller "download" or similar icon in a corner of an image, maybe hidden until image is hovered.
> 
> If we do bake that into images in general, shouldn't be too hard to override that behavior if we want to just have a straight up image link or whatever.

I see no problem in adding a wrapper component for ResponsiveImage with extra functionality. Would be quite easy to define a `downloadable` boolean and other settings for the component.

A bit out of the scope for this though, but I think it would be nice to have. The images we have and store are quite functional, and I think we can use them in more ways.

> I did something about this whole thing when working on the employee list for Bekk last summer I think, but that might have been just not having a static "page" width and instead adapting that too. That could be an interesting idea, probably hard to get the footer to align too, but at least the page content could be synchronized potentially. No idea, don't really have time to look at it, but I suppose we'll both have access to the employee list come Monday, which is probably way improved from when I worked on it a year ago. Should be okay to open up the inspect tool and see how it's done.

Interesting, I will have to look into that on Monday :P

Do you think the current implementation works OK enough to be used though?

plusk:
This PR is about designing new hobby cards, and it looks like that's been done. Making better responsiveness is more of a site-wide overhaul anyways

src/common/less/constants.less
plusk: watch your language young man
but yeah, is this gonna leave some 👉👌🆙 CSS @oleast?

PR #304, opened 2019-10-02
hides committees if not in committee

### LINKED ISSUE #279: Sections in profile should be hidden if they have no information to show
This includes, but is not limited to:
eksterne sider
komite daljer
Komité-e-post?

An alternative to hiding it would be to add some text, encouraging you to participate, for it to eventually show up

VigdisIS:
fixes #279 

Checks if user is in a committee and displays committee medals thereafter.

### DIFF
--- a/src/profile/components/Profile/MainProfile.tsx
+++ b/src/profile/components/Profile/MainProfile.tsx
@@ -28,13 +28,13 @@ export const MainProfile = () => {
         <Study />
       </SplitPane>
       <Bio />
-      {profile.positions.length > 0 ?
+      {profile.positions.length > 0 ? (
         <Pane>
           <Content title="Komitéverv">
             <MedalsView medals={profile.positions} />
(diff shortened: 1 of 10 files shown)


src/profile/components/Profile/MainProfile.tsx
```diff
@@ -28,11 +28,13 @@ export const MainProfile = () => {
...
-          <MedalsView medals={profile.positions} />
-        </Content>
-      </Pane>
+      {profile.positions.length > 0 ?
+        <Pane>
+          <Content title="Komitéverv">
+            <MedalsView medals={profile.positions} />
+          </Content>
+        </Pane>
+        : <></>}
```
henrikhorluck: You don't need to have a fragment here, null is sufficient

VigdisIS: Fixed.

PR #309, opened 2019-10-09
Dependency upgrade

henrikhorluck:


package.json
```diff
@@ -134,9 +133,9 @@
     "react-helmet-async": "^1.0.2",
     "react-loadable": "^5.5.0",
     "react-redux": "^7.1.0",
-    "react-select": "^2.4.2",
+    "react-select": "^3.0.8",
     "react-spring": "^8.0.20",
-    "react-stripe-elements": "^3.0.0",
+    "react-stripe-elements": "^5.0.1",
```
henrikhorluck: As one can se from https://github.com/stripe/react-stripe-elements/releases
These are the breaking changes:
> We replaced the internal use of deprecated componentWillReceiveProps. This internal movement of logic between lifecycle methods is likely safe for almost all apps and should not require any changes.

> CardCVCElement has been renamed to CardCvcElement. CardCVCElement will
be removed in version 5.0.0.

> If you were already using handleCardSetup with react-stripe-elements, you
should upgrade your integration. This method will now automatically find and
use valid Elements.

Neither of which affects us as far as I know.

types/react-helmet-async.d.ts
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  // tslint:disable-next-line:interface-name
+  interface ProviderProps {
+    context?: {};
+  }
+
+  // @ts-ignore
+  export class HelmetProvider extends React.Component<ProviderProps> {}
+}
```
henrikhorluck: This was needed because the typings were somehow out of sync, and was giving me
```
TS2604: JSX element type 'Helmet' does not have any construct or call signatures.
```

src/articles/components/ArticleView/index.tsx
```diff
@@ -37,6 +37,8 @@ export const ArticleView = ({ articleId }: IProps) => {
 
   return (
     <div className={style.container}>
+      {/*
+      // @ts-ignore-next-line TS2604 */}
```
henrikhorluck: I can confirm this still works, even in SSR, but Typescript is not happy for some reason, so this was added, it is likely we can remove it at some point in the near future, but unsure about what caused it in the first place.

Issue #252, opened 2019-05-19
Let user edit their password


oleast:
Users should be able to change their own password through the profile page.

PR #311, opened 2019-10-10
Added settings functionality to change your password

kharann:
You're now able to change you user password on OWF :tada: 

Here are some images

### How the change password form usually look
editPassword1
### How it looks when you have not typed 8 characters (8 is min for our passwords)
editPassword2
### How it looks when you submit with wrong current password or retype the password wrong
editPassword3

## Issues
So what issues do we have? Guess what i had to wrap my PUT request inside of a try catch because when the password successfully changes it doesnt return anything, which means the PUT request tries to make nothing into json === error! Our PUT request expects a response, usually the data we sent. However responding with the password you just typed is pretty stupid. What we have to do is to rewrite our PUT request. This error when it returns nothing makes it also a bit difficult for us to display a message that the user sucessfully changed their password because we dont know if the "try catch" catches the empty response, or an actual error.

### DIFF
--- a/src/common/utils/api.ts
+++ b/src/common/utils/api.ts
@@ -40,6 +40,9 @@ const performRequest = async (query: string, parameters: IQueryObject = {}, opti
   };
   const requestOptions = { ...restOptions, headers };
   const response = await fetch(url, requestOptions);
+  if (response.status === 204) {
+    return null;
+  }
   const data = await response.json();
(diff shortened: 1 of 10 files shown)


kharann:
How it looks if you successfully change your mail
nice

kharann:
fixes #252

PR #313, opened 2019-10-10
Style input-type search correctly for Safari

henrikhorluck:
Because Safari styles the `<input type="search">` element differently
How it currently looks in Safari:
image
How it looks with this change:
image

This makes no functional difference

### DIFF
--- a/src/common/components/Forms/SearchInput/search.less
+++ b/src/common/components/Forms/SearchInput/search.less
@@ -20,3 +20,7 @@
     background: darken(@pageBackground, 5%);
   }
 }
+
+input[type="search"] {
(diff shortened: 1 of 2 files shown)


oleast:
This is just a problem in terms of styling. Doing something like:

``` css
input[type="search"] {
  -webkit-appearance: textfield;
}
```

Should let you style the field correctly in Safari (...).

There is a functional difference in functionality for type="text" and type="search" in terms of accessibility, which should not be ignored.

henrikhorluck:
Thanks for the input! Changed the PR accordingly, and I can confirm that that does indeed work

PR #316, opened 2019-10-16
Adds contact info to footer

### LINKED ISSUE #250: Add missing contact information
With the footer recently having gone on a diet (#249), OWF is missing some contact information that could be useful for companies, new or existing students. We have a couple of alternatives:

Putting the information in fuzzy-train would make the information immediately available to companies (and new students, should they need any of it). However, then we have to find a place for it in fuzzy-train. The information might be a bit harder to get to if you're an existing student, however.

As an alternative (or an addition), we could have a static contact page on OWF. Just a few boxes of information, add in a Google Map or the likes if you're feeling cheeky (there's really no need). If contact information was added to chunks in some way, then we could just have a page pulling from some of the same chunks as fuzzy-train.

In both cases we'll run into the issue of "where do we link to the contact information?", as we haven't quite figured out how we want to link back to the splash page yet. The header is pretty filled up, and it doesn't really fit the new footer either, which would be a problem for a static contact page on OWF too.

R33an1:
Resolves #250.

Adds contact info to the footer. 

More than happy to get some CC.

image

### DIFF
--- a/src/core/components/Footer/ContactInfo.tsx
+++ b/src/core/components/Footer/ContactInfo.tsx
@@ -7,9 +7,15 @@ const ContactInfo = () => (
   <div className={style.contact}>
     <div>
       <FontAwesomeIcon icon={faBriefcase} fixedWidth className={style.icon} /> 992 548 045 (OrgNr) <br />
-      <FontAwesomeIcon icon={faEnvelope} fixedWidth className={style.icon} /> kontakt@online.ntu.no <br />
-      <FontAwesomeIcon icon={faFile} fixedWidth className={style.icon} /> okonomi@online.ntnu.no <br />
-      <FontAwesomeIcon icon={faMobileAlt} fixedWidth className={style.icon} /> 73 59 64 89
+      <a href="mailto:kontakt@online.ntu.no">
+        <FontAwesomeIcon icon={faEnvelope} fixedWidth className={style.icon} /> kontakt@online.ntu.no <br />
+      </a>
+      <a href="mailto:okonomi@online.ntnu.no">
+        <FontAwesomeIcon icon={faFile} fixedWidth className={style.icon} /> okonomi@online.ntnu.no <br />
+      </a>
+      <a href="tel:+4773596489">
+        <FontAwesomeIcon icon={faMobileAlt} fixedWidth className={style.icon} /> 73 59 64 89
+      </a>
     </div>
     <div>
       Besøksadresse: <br />
--- a/src/core/components/Footer/ContactInfo.tsx
+++ b/src/core/components/Footer/ContactInfo.tsx
@@ -5,12 +5,12 @@ import style from './footer.less';
 
 const ContactInfo = () => (
   <>
-    <div>
+    <address>
       <p>Besøksadresse: </p>
       <p>A-blokka, A4-137</p>
       <p>Høgskoleringen 5 </p>
       <p>NTNU Gløshaugen</p>
-    </div>
+    </address>
     <div>
       <p>Kontaktinformasjon:</p>
       <p>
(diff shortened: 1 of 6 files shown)


henrikhorluck:
Some images? 📸

src/core/components/Footer/ContactInfo.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const ContactInfo = () => (
+  <div className={style.contact}>
+    <div>
+      <FontAwesomeIcon icon={faBriefcase} fixedWidth className={style.icon} /> 992 548 045 (OrgNr) <br />
+      <FontAwesomeIcon icon={faEnvelope} fixedWidth className={style.icon} /> kontakt@online.ntu.no <br />
+      <FontAwesomeIcon icon={faFile} fixedWidth className={style.icon} /> okonomi@online.ntnu.no <br />
+      <FontAwesomeIcon icon={faMobileAlt} fixedWidth className={style.icon} /> 73 59 64 89
+    </div>
+    <div>
```
henrikhorluck: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/address

R33an1: Will fix, what kind of images? :P

henrikhorluck: How does the footer look like after the change?

henrikhorluck: It is a bit too much effort for me to run it locally just to check a small design change 😅

R33an1: Added a screenshot. Totally forgot to do that :P

henrikhorluck:
Also, are we certain we should solve this by just adding it to the footer, and not creating a new page for stuff like this and list of commitee-members and such?

R33an1:
> Also, are we certain we should solve this by just adding it to the footer, and not creating a new page for stuff like this and list of commitee-members and such?

As I understood it we wanted to have somewhere companies can quickly find this information without having to dig through the site. That being said I'm sure we could have less information in the footer and rather have the remaining bits somewhere else.

plusk:
Aesthetically this looks perfectly fine, only nitpick is that it would be cool if the icons had the same dark gray color instead of black. 
As far as "should this be here?" goes, yeah maybe. The whole company information part of things has been put on hold until we decide what to do with fuzzy train, but I'm leaning towards the OW4 approach. As part of that we'd need this information somewhere easily accessible too.
Practically speaking, I'd say this can be merged. If we decide on a different solution, it won't be hard to remove. Temporary solutions are totally valid, and this might be the permanent solution anyways.

oleast:
From a design standpoint, I think it would help to have both rows be the same width, as well as make all icons in the footer have the same color to not stand as much out?

R33an1:
> From a design standpoint, I think it would help to have both rows be the same width, as well as make all icons in the footer have the same color to not stand as much out?

I'll take a look at making them the same width, the color however is #333333 for both the text and the icons.

FluidSense:
I picked this up again, as it seemed stale!
I've tried to place all items at same width, and scale to screen size.

src/core/components/Footer/index.tsx
```diff
@@ -1,19 +1,23 @@
...
-      <Img src="/static/img/hsp-832576.svg" alt="Hovedsamarbeidspartner - Bekk" />
-    </a>
-    <p>
-      Har du funnet en feil på nettsiden?
-      <br />
-      Ta kontakt med <a href="mailto:dotkom@online.ntnu.no">Utviklingsteamet</a>
-    </p>
-    <SocialLinks />
+    <div className={style.footerContent}>
+      <a href="http://www.bekk.no/" className={style.sponsor}>
```
henrikhorluck, line 10:
```suggestion
      <a href="https://www.bekk.no/" className={style.sponsor}>
```

PR #321, opened 2019-10-23
Minor-Refactor password and add autocomplete attributes

henrikhorluck:
Example of what your password manager can do now:
Screenshot 2019-10-23 at 13 19 05

Note I did remove explicit `return`-s that can be implicit, otherwise check commit messages for changes


src/profile/components/Settings/Password/ErrorMessage.tsx
```diff
@@ -5,16 +5,14 @@ interface IProps {
...
+  <>
+    {errors
+      ? errors.map((error) => (
+          <li className={style.errorMessage}>
+            <ol>{error}</ol>
+          </li>
+        ))
+      : null}
+  </>
+);
```
oleast: There is really no real reason to do this?

If we were to change something I would rather just remove the fragments:
``` typescript
export const ErrorMessage: FC<IProps> = ({ errors }) => {
  return errors
    ? errors.map((error) => (
        <li className={style.errorMessage}>
          <ol>{error}</ol>
        </li>
      ))
    : null;
};
// Or some formatting like that I guess
```

src/profile/components/Settings/Password/PasswordInput.tsx
```diff
@@ -1,23 +1,20 @@
...
-};
+const PasswordInput: FC<IProps> = ({ label, requiredMessage, ...props }) => (
+  <div className={style.passwordInput}>
+    <label>
+      {label}
+      <input {...props} type="password" minLength={8} />
+    </label>
+    <ErrorMessage errors={requiredMessage} />
+  </div>
+);
```
oleast: No real reason to remove the return, it just makes it harder to add hooks to the component.

src/profile/components/Settings/Password/SuccessMessage.tsx
```diff
@@ -6,8 +6,8 @@ interface IProps {
   message: string;
 }
 
-const SuccessMessage: FC<IProps> = ({ success, message }) => {
-  return <>{success && <div className={style.successMessage}>{message}</div>}</>;
-};
+const SuccessMessage: FC<IProps> = ({ success, message }) => (
+  <>{success && <div className={style.successMessage}>{message}</div>}</>
+);
```
oleast: Same as the others.

PR #326, opened 2019-11-06
Fix firefox frontpage bug for the event containers

kharann:
Pretty weird bug at least for my firefox. Pretty much a very simple fix

Before on firefox:
wtf
After on firefox:
wtf2


FluidSense:
@kharann Is this still a problem? It seems fine for me on current prod with Firefox 75

FluidSense:
I'll close this as stale.
Feel free to reopen if the problem persists and this is a proposed fix.

PR #331, opened 2020-01-27
Quick tweaks that slipped through the cracks

plusk:
- Adjusted headed color to just be our new primary blue color, instead of a darker version
- Tweaked calendar for tablet and mobile, with a new hover on tiles
- Added hover for image event views

I think larger design changes should be handled when we integrate DS instead

### DIFF
--- a/src/core/components/Header/header.less
+++ b/src/core/components/Header/header.less
@@ -6,7 +6,7 @@
 @contentHeight: @headerHeight - @headerPaddingVertical * 2;
 
 .header {
-  background: @darkBlue;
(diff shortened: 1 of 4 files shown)


oleast:
I think they are still relevant

I had a look at them when the PR was set up, but took the comments in person.
There seems to bee scaling problems for the calendar for some screen sizes, I rebased the branch now, so we'll get a preview in a sec

oleast:
Can't see any of the problems which others than already existing ones, I say merge :heart:

PR #333, opened 2020-03-22
Update webpack to properly run dev

FluidSense:
Webpack 4.29 has an issue which caused react-loader to fail. This is seemingly fixed in 4.42


henrikhorluck:
Perhaps regenerate `yarn.lock` and rebase on master for CI to work?

PR #334, opened 2020-04-01
Feat/oidc client managment

Aborysa:
Created interface for the user to download their data, anonymize their account, see their client connections and manage their apps.

Some changes will have to be done to the backend models before this pr can be merged (will create pr asap).


src/profile/components/Settings/UserData/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+      return () => {
+        URL.revokeObjectURL(ref);
+        setFileRef(null);
+      };
+    }
+    return;
+  }, [userData]);
+
+  useEffect(() => {
+    if (aRef.current != null) {
```
oleast, line 58: `!==`?

You're already checking if it's `null`, why the ts-ignore?

src/profile/components/Settings/UserData/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+import style from './userdata.less';
+
+interface IUserCredentials {
+  username: string;
+  password: string;
+}
+
+export const UserData = () => {
+  const userContext = React.useContext<IUserContext>(UserContext);
+  const user = userContext.user;
```
oleast, line 17:
``` typescript
const { user } = React.useContext<IUserContext>(UserContext);
```
Maybe a bit simpler like this?

src/oidc_clients/models/client.ts
```diff
@@ -0,0 +1,25 @@
+import { IResponseType } from './responsetype';
+
+export interface IOidcClientOwner {
+  first_name: string;
+  last_name: string;
+  email: string;
+  username: string;
+}
```
oleast, line 8: I think we already have this defined as a type, something like `IPublicUser` or something like that?

src/oidc_clients/components/OidcClient/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const clientDescriptionMd = `
+  # ${client.name}
+  - Client id: ${client.client_id}
+  - Eier: ${client.owner ? client.owner.first_name : 'Ingen'}
+  - Tilganger (scope): ${client.scope.length > 0 ? client.scope.join(', ') : 'Ingen'}
+  - Website url: ${client.website_url}
+  - Client type: ${client.client_type}
+  - Response types: ${client.response_types.map((rt) => rt.description).join(', ')}
+
+`;
```
oleast, line 22: Could just use the `md` template-tag exported from the same file as the `Markdown` component?

src/oidc_clients/api/user_consent.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const getUserConsent = async (user: IAuthUser) => {
+  const consents: IUserConsent[] = [];
+  const consentResponse = await getAllPages<ConsentResponseType>(API_BASE + 'consent/', undefined, { user });
+
+  for (const consent of consentResponse) {
+    const client = await getClient(consent.client as number);
+    consent.client = client;
+    consents.push(consent as IUserConsent);
+  }
```
oleast, line 20: This would still be a synchronous loop, would it not?
We could speed it up a bit by mapping a `Promise.all`?

src/index.tsx
```diff
@@ -20,6 +22,14 @@ import App from './App';
...
 registerServiceWorker();
 
+// test
+
+/*import { getMyClients } from './oidc_clients';
+
+getMyClients().then((...a) => {
+  console.log(...a);
+});*/
+
```
oleast, line 32: Could delete this

src/common/utils/api.ts
```diff
@@ -107,7 +107,26 @@ export const post = async <T>(
     'Content-Type': 'application/json',
   };
   const body = JSON.stringify(data);
-  const opts = { ...options, method: 'POST', body, headers };
+  const opts = { ...options, method: 'DELETE', body, headers };
```
oleast, line 110: POST is changed to DELETE?

package.json
```diff
@@ -106,28 +105,35 @@
...
     "abortcontroller-polyfill": "^1.2.3",
     "assets-webpack-plugin": "^3.9.7",
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "^24.1.0",
     "classnames": "^2.2.5",
+    "core-js": "^3.6.4",
     "express-http-proxy": "^1.5.0",
     "focus-visible": "^4.1.5",
     "isomorphic-fetch": "^2.2.1",
+    "markdown-to-jsx": "^6.11.0",
```
oleast, line 136: We already have a markdown lib, why do we need another?

package.json
```diff
@@ -106,28 +105,35 @@
...
   "dependencies": {
     "@babel/plugin-syntax-dynamic-import": "^7.2.0",
+    "@dotkomonline/design-system": "^0.14.0",
+    "@dotkomonline/react-components": "^2.0.0",
     "@fortawesome/fontawesome-svg-core": "^1.2.14",
     "@fortawesome/free-brands-svg-icons": "^5.8.2",
     "@fortawesome/free-regular-svg-icons": "^5.7.2",
     "@fortawesome/free-solid-svg-icons": "^5.7.1",
     "@fortawesome/react-fontawesome": "^0.1.4",
+    "@mdx-js/loader": "^1.5.7",
```
oleast, line 115: We don't have MDX in this project, should not need to install this?

package.json
```diff
@@ -106,28 +105,35 @@
   },
   "dependencies": {
     "@babel/plugin-syntax-dynamic-import": "^7.2.0",
+    "@dotkomonline/design-system": "^0.14.0",
+    "@dotkomonline/react-components": "^2.0.0",
```
oleast, line 109: This is not needed anymore?

.babelrc
```diff
@@ -3,12 +3,9 @@
...
     "@babel/react",
     ["@babel/env", {
-      "targets": {
-        "chrome": "64"
-      },
-      "useBuiltIns": "usage",
       "debug": false,
-      "corejs": 2
+      "corejs": 3,
+      "useBuiltIns": "entry"
```
oleast, line 8: This may break polyfills for `Array.prototype.flat/flatMap` , which is hard to notice.
Or maybe not, but bad if it happens again

.babelrc
```diff
@@ -17,6 +14,6 @@
     "@babel/proposal-optional-chaining",
     "@babel/plugin-syntax-dynamic-import",
     ["@babel/plugin-proposal-decorators", { "legacy": true }],
-    ["@babel/plugin-proposal-class-properties", { "loose": true }]
+    ["@babel/plugin-proposal-class-properties"]
```
oleast, line 17: What is the end result of this?

src/oidc_clients/api/clients.ts
```diff
@@ -0,0 +1,70 @@
...
+// fetch clients owned by me
+// todo: filter matches on username, client.owner should probably contain user id aswell
+export const getMyClients = async (user: IAuthUser): Promise<IOidcClient[]> => {
+  if (!user) {
+    return [];
+  }
+  const clients = await getAllClients();
+  const filtered = clients.filter((client) => {
+    return client.owner && client.owner.username === user.profile.preferred_username;
+  });
```
oleast, line 63: Could we filter this in the query to the API instead?

src/profile/components/Settings/Apps/AppConnections.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export const AppConnections = () => {
+  const userContext = useContext<IUserContext>(UserContext);
+  const user = userContext.user;
+
+  if (!user) {
+    return <></>;
+  }
+
+  const [removeConsentState, setRemoveConsentState] = React.useState<IRemoveState>({
```
oleast, line 27: Would go for importing `useState` since that is done all other places in the project

src/profile/components/Settings/Apps/AppConnections.tsx
```diff
@@ -0,0 +1,75 @@
...
+import style from './apps.less';
+
+interface IRemoveState {
+  client_id: number;
+  client?: IOidcClient;
+}
+
+export const AppConnections = () => {
+  const userContext = useContext<IUserContext>(UserContext);
+  const user = userContext.user;
```
oleast, line 21: Destructure here as well?

src/profile/components/Settings/Apps/AppConnections.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }, [removeConsentState, state1, revokeState]);
+
+  if (state1 === 'resolved' && state1 === 'resolved') {
+    consentedApps = consents.map((consent: IUserConsent) => {
+      return (
+        <li key={consent.id}>
+          <OidcClient
+            onRemove={() => {
+              setRemoveConsentState({ client_id: consent.id, client: consent.client });
+              // deleteConsentCall(user, consent.id);
```
oleast, line 55: Remove this line?

src/profile/components/Settings/Apps/ManageApps.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const ManageApps = () => {
+  const userContext = useContext<IUserContext>(UserContext);
+  const user = userContext.user;
+
+  if (!user) {
+    return <></>;
+  }
+
+  const [removeClientState, setRemoveClientState] = React.useState<IRemoveState>({
```
oleast, line 25: Same here, import `useState`

FluidSense:
@Aborysa Cool stuff! Could you add some screenshots of the new features from the user perspective? Or a pointer towards where to look when testing?

src/profile/components/Settings/UserData/index.tsx
Aborysa: the linter would still report it as a warning even tho I wrapped it in a null check.

src/oidc_clients/models/client.ts
Aborysa: Ah, I'll look into that. Thanks

src/common/utils/api.ts
Aborysa: Oops

package.json
Aborysa: There was some issues with importing design system, webpack would refuse to compile without adding markdown-to-jsx to the project. I think this might be solved by adding markdown-to-jsx in designsystems dependencies instead of dev-dependencies.

Aborysa:
@FluidSense will do

src/oidc_clients/api/clients.ts
Aborysa: How would I approach that? I have tested using filter "name=nibble-cli" and "owner.username=jannetr" but it doesn't actually filter out anything.

src/profile/components/Settings/Apps/AppConnections.tsx
Aborysa: :+1:

src/oidc_clients/models/client.ts
Aborysa: The user types we have in profile/models/User doesn't overlap properly with IOidcClientOwner.

Aborysa:
Some pictures

image

image

src/oidc_clients/models/client.ts
FluidSense: Is there too many fields or is something lacking?
`interface IOidcCliendOwner extends Partial<IPublicUser>` could perhaps be used then, to symbolize a relation between the models?

package.json
```diff
@@ -8,9 +8,9 @@
   "private": true,
   "type-check": "tsc --jsx",
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 8080",
```
oleast, line 11: This was changed to `3000` to keep with the defaults

src/common/utils/api.ts
```diff
@@ -43,6 +43,11 @@ const performRequest = async (query: string, parameters: IQueryObject = {}, opti
   if (response.status === 204) {
     return null;
   }
+  // encountered a http error, throw error
+  if (response.status > 399) {
+    throw response;
+  }
```
oleast, line 49: Are we sure this doesn't break any existing code?

src/common/utils/api.ts
```diff
@@ -73,20 +78,24 @@ export async function getAllPages<T>(
...
+      get<IAPIData<T>>(query, { ...parameters, page: i + 1, page_size }, options)
+    );
+    /** Await all the fetches to a single array */
+    const data: Array<IAPIData<T>> = await Promise.all(requests);
+    /** Reduce all results to a single array for all objects in the resource */
+    const results = data.reduce<T[]>((res, d) => res.concat(d.results), []);
+    return results;
+  } catch (response) {
+    return [];
+  }
```
oleast, line 98: Should this fail silently?

src/core/appUrls.ts
```diff
@@ -24,7 +24,11 @@ export const getProfileSettingsNotificationsUrl = () => url`/profile/settings/no
 export const getProfileSettingsPasswordUrl = () => url`/profile/settings/password`;
 export const getProfileSettingsPenaltiesUrl = () => url`/profile/settings/penalties`;
 export const getProfileSettingsPrivacyUrl = () => url`/profile/settings/privacy`;
+export const getProfileAppConnectionsUrl = () => url`/profile/settings/apps`;
+export const getProfileUserDataUrl = () => url`/profile/settings/userdata`;
+
```
oleast, line 29: Could we remove the spaces?

src/oidc_clients/api/clients.ts
oleast: We would have to add it in the back-end, could look into it for future stuff

src/pages/_app.tsx
```diff
@@ -34,17 +36,20 @@ type Props = AppProps & ReduxWrapperAppProps<State>;
...
-          <UserProfileProvider>
-            <Core>
-              <Component {...pageProps} />
-            </Core>
-          </UserProfileProvider>
-        </AuthProvider>
-      </ContextWrapper>
-    </Provider>
+    <>
+      <GlobalStyle />
```
oleast, line 40: Adding these styled requires additions in `_document.tsx` to have the CSS for initial page load with SSR

Example of it can be found here: https://github.com/dotkom/gallery/blob/master/src/pages/_document.tsx#L5-L29

src/profile/components/Settings/UserData/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import style from './userdata.less';
+
+interface IUserCredentials {
+  username: string;
+  password: string;
+}
+
+const UserData = () => {
+  const userContext = React.useContext<IUserContext>(UserContext);
```
oleast, line 16: `React.useContext<IUserContext>(UserContext)` -> `useContext(UserContext);` ?

src/oidc_clients/models/client.ts
oleast: We apparently have two serializers for displaying users, we should probably just have one: https://github.com/dotkom/onlineweb4/blob/develop/apps/authentication/serializers.py#L27-L39

src/common/utils/api.ts
Aborysa: `...Array(pageNumber)` fails with `Invalid array length` if  `await get<IAPIData<T>>(query, { ...parameters, page, page_size: 1 }, options)` returns a 404, because `undefined / page_size` give NaN. Should probably revise the getAllPages api in the future, but I'll leave it like this for now.

src/common/utils/api.ts
Aborysa: Changed it to only throw the response when response.json fails and when a 403 or 5xx error happens. This should not break any existing apis. Throwing a error on 403 and 5xx is required by the oidc api in order to show the user an error when their token has expired(403) and when data-dump and anonymize fails due to internal server error.

package.json
Aborysa: 3000 conflicts with webpack on onlineweb4 when running locally

FluidSense: Should this perhaps be added to the README then?

Aborysa: Can revert it and make the change in another pr. But yes, should be included in the README

src/common/hooks/useAsync.ts
```diff
@@ -0,0 +1,156 @@
...
+      };
+    case 'rejected':
+      return {
+        result: undefined,
+        error: action.value,
+        status: 'rejected',
+      };
+  }
+  return state;
+};
```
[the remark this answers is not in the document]
Aborysa: Not sure I understand what you mean?

PR #335, opened 2020-04-02
Remove host volumes for cloud usage in .drone.yml

FluidSense:


FluidSense:
9d8ecbd borked cache, oops.
AWS S3 such as @tokongs has made in https://github.com/dotkom/onlineweb4/pull/2461 is probably the way to go.

oleast:
Looks the linked commit has been merged.
Would it be possible to copy the configuration from that project? @FluidSense 😊

FluidSense:
@oleast The cache part, yes! The rest is a bit different since this is TypeScript and no backend ofc

.drone.yml
```diff
@@ -98,6 +130,7 @@ steps:
       command_timeout: 600s
       script:
         - /srv/www/owf-staging/deploy.sh
+    depends_on: [ clone ]
```
oleast: I think maybe deployment should depend on the project at least compiling correctly?

FluidSense: Yeah, I was first under the impression that this dependency "cloned" all dependencies from the last job, but as I found no documentation of it, I added all steps manually.

PR #336, opened 2020-04-02
Refactor project to NextJS

oleast:
This project has become _very_ complicated, a lot more than it needs to be.
In an effort to remove some of that complexity from development and deployment I'm trying to change the structure to NextJS.

Next provides a lot of the custom implemented features of the server-side rendering back-end we currently have, but without us having to maintain them ourselves, and with even more features. The project is extremely well maintained, and has new features incoming all the time. The community is huge, and you'll find fixes for your problems when you search for them.

This is going to be a large, multi-step undertaking. I'll try to implement the basic things first, just to make everything _work_. Then we'll go over to getting other features like Data fetching on the server, static pages and the next router to work completely.


next.config.js
```diff
@@ -0,0 +1,40 @@
+const withLess = require('@zeit/next-less');
+const withCss = require('@zeit/next-css');
+const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
+
+module.exports = withCss(
```
henrikhorluck, line 5: `nextCss` isn't necessary anymore as of next 9.2

oleast: That is for pure CSS, and SCSS if i recall correctly.

It is still needed for LESS support: https://github.com/zeit/next-plugins/issues/598

oleast:
Resolved by one of the subsequent PRs based on this

PR #338, opened 2020-04-05
Show today #314

### LINKED ISSUE #314: No visual clue to know which date it is in event calendar view
When viewing events in the calendar view there is no visual clue (or other clues) to know what date it currently is

Shading the current day's background to gray or some other clue

kaffikaffi:
Resolves #314 

I have added basic functionality for showing which day it is on the calendar, I think it looks nice, and makes it easier to use the calendar :dancing_men: 

image

### DIFF
--- a/src/events/components/CalendarView/CalendarTile.tsx
+++ b/src/events/components/CalendarView/CalendarTile.tsx
@@ -1,8 +1,7 @@
 import classNames from 'classnames';
-import { DateTime } from 'luxon';
 import React, { FC } from 'react';
 import { shallowEqual } from 'react-redux';
-
+import { DateTime } from 'luxon';
 import { eventSelectors } from 'events/slices/events';
 import { useSelector } from 'core/redux/hooks';
 import { State } from 'core/redux/Store';
@@ -17,6 +16,7 @@ interface IProps {
 
 export const EventCalendarTile: FC<IProps> = ({ date, active }) => {
   const eventIds = useSelector(selectEventIdsForDate(date), shallowEqual);
+  console.log(date);
   return (
     <div
       className={classNames(style.tile, {
@@ -24,7 +24,13 @@ export const EventCalendarTile: FC<IProps> = ({ date, active }) => {
       })}
     >
       <div className={style.tileContent}>
-        <p>{date.day}</p>
+        <p>
+          {isSameDate(date, DateTime.local()) ? (
+            <span className={classNames(style.tileToday)}>{date.day}</span>
+          ) : (
+            date.day
+          )}
+        </p>
         {eventIds.map((eventId) => (
           <CalendarEvent key={eventId} eventId={eventId} />
         ))}
--- a/src/events/components/CalendarView/CalendarTile.less
+++ b/src/events/components/CalendarView/CalendarTile.less
@@ -22,7 +22,17 @@
   background: @offWhite;
   box-shadow: none;
 }
-
+.tileToday {
+  display: block;
+  text-align: center;
+  padding: 3px;
+  width: 1rem;
+  height: 1rem;
+  color: @offWhite;
+  background-color: @darkBlue;
+  border-radius: 50%;
+  line-height: 1;
+}
 .tileContent {
   padding: 0.25rem;
 


src/events/components/CalendarView/calendar.less
```diff
@@ -26,6 +26,20 @@
   height: 100px;
   background: #fff;
 }
+.tileToday {
+  &:extend(.tile);
```
oleast, line 30: Since the html element already has the class "tile" I think this would just lead to duplicated CSS?

src/events/components/CalendarView/index.tsx
```diff
@@ -76,7 +76,13 @@ export const CalendarView = () => {
         <CalendarFillerTiles days={previous} />
         {displayEventMonth &&
           displayEventMonth.map((events, index) => (
-            <CalendarTile key={`${month.toFormat('yyyy-MM')}-${index + 1}`} events={events} day={index + 1} active />
+            <CalendarTile
+              key={`${month.toFormat('yyyy-MM')}-${index + 1}`}
+              events={events}
+              day={index + 1}
+              month={month}
```
oleast, line 83: I think it is quite strange that our CalendarTile component doesn't know which date it is displaying, like as a whole date object. I would blame myself for that.

What if we removed the `day: number` and `month: DateTime` props, and just replace it with a `date: DateTime` prop which is the actual year, month and day the Tile is supposed to represent?

I think something like this would work, but that is without testing it:
``` typescript
// Taking the month object, and setting the day to be the day of the tile
const date = month.set({ day: index + 1 })

// How it could look in the JSX
<CalendarTile
  key={`${month.toFormat('yyyy-MM')}-${index + 1}`}
  events={events}
  date={month.set({ day: index + 1 })}
  active
/>
```

This could also make the `isToday` comparison a lot simpler.
We could possibly just do something like:
``` typescript
  // Get the current date as a Luxon DateTime
  const today = DateTime.local();
  // Compare just the string representation of the year, month and day
  const isToday = date.toFormat('yyyy-MM-dd') === today.toFormat('yyyy-MM-dd');
```

src/events/components/CalendarView/CalendarTile.tsx
```diff
@@ -20,15 +22,24 @@ export const createDayList = (amount: number, start: number): number[] => {
   return l;
 };
 
-export const CalendarEventTile = ({ events, active = true, day }: ITileProps) => {
+export const CalendarEventTile = ({ events, active = true, day, month }: ITileProps) => {
+  const thisDate = new Date();
+  const today = thisDate.getDate();
+  const thisMonth = thisDate.getMonth() + 1;
+  // tslint:disable-next-line: radix
```
oleast, line 29: The Radix is a required parameter of the `parseInt` function, or at least it is according to tslint.
The idea is that `parseInt` can parse other number systems than just base 10. `parseInt('11', 10) would give you `11`, while`parseInt('11', 2) would give you `3` because binary. Tslint is just trying to guard against errors here I think

src/events/components/CalendarView/CalendarTile.tsx
```diff
@@ -20,15 +22,24 @@ export const createDayList = (amount: number, start: number): number[] => {
   return l;
 };
 
-export const CalendarEventTile = ({ events, active = true, day }: ITileProps) => {
+export const CalendarEventTile = ({ events, active = true, day, month }: ITileProps) => {
+  const thisDate = new Date();
+  const today = thisDate.getDate();
+  const thisMonth = thisDate.getMonth() + 1;
+  // tslint:disable-next-line: radix
+  const isToday = day === today && thisMonth === parseInt(month.toFormat('M')) ? true : false;
```
oleast, line 30: I don't think this would take the year into account, so it could possibly turn a false positive if things line up correctly?

oleast, line 30: When it comes to the ternary statement we can drop the `? true : false` part.

Example:
``` typescript
// In this case `isToday` would evaluate to a boolean, so adding the ternary would have no effect.
const isToday = day === today && thisMonth === parseInt(month.toFormat('M'));

// If we were to use it for something like a text to display, the ternary could be very useful
const dayText = day === today && thisMonth === parseInt(month.toFormat('M')) ? 'Today is the day' : 'Today is not the day'
```

src/events/components/CalendarView/index.tsx
FluidSense: This comment is fixed in #360, is it not @oleast?

oleast: Yes :smile:

aslakhol (approved):
Resolve the merge conflict and then we are ok. We added issue #376 to handle the problem Ole described.

kaffikaffi:
Looks kinda shit for now, but i will fix :)

kharann:
yep yep, looks good, yep yep

PR #341, opened 2020-04-05
Upgrade ts version to 3.8.3

FluidSense:
This spots a few type-error the previous version did not spot.
This PR fixes the same errors.

The fix changes the conditional rendering of Payment, which no longer wraps the error inside the Payment component. This should perhaps be double-checked that looks correctly.


FluidSense:
Merging although Now-bot can't deploy. I think that is added support for in a newer PR.

Issue #320, opened 2019-10-22
Event DetailView doesn't show markdown in the same way as OW4


henrikhorluck:
**Describe the bug**
Look at https://beta.online.ntnu.no/events/785, as you can see, there is multiple places with `######a new sentence`, when that should be rendered as a `<h6/>`-tag
Example on OW4 https://online.ntnu.no/events/785/tdc-aspire-2019/

**Additional context**
This has also been a problem with articles, but was solved by using a regex-replace, that ensured there was a space between the last `#` and the start of the next sentence. While we could use that here, at this point it would probably be better to look into if we can config `<Markdown/>` to parse the markdown as wanted

oleast:
This problem is solved by just writing markdown correctly.
We should not have to fix the code because of wrong markdown.

henrikhorluck:
But is _is_ correct when looking at OW4, and I think is a dumb breaking change, since we need to change many earlier versions  of articles and events

FluidSense:
I think I agree with @oleast on this one.
Although OW4 has been forgiving so far, OWF does not need to allow wrong markdown. 
People will learn, people will fix, and if people view earlier versions, it's not that big of an issue if they look a little weird.
I'm inclined to close this issue as a `wontfix` unless there are any more arguments as to why we should allow wrongly-written markdown.

FluidSense:
I will close this issue, feel free to reopen if there are more arguments or multiple users asking for the feature!

PR #342, opened 2020-04-07
Refactor routing to Next Router

oleast:
Makes use of the file-system based routing in NextJS

This has some advantages, and some disadvantages:

The main things are that it is a bit easier to reason about.
Each file is an entrypoint for a route, request an URL and the default export of that file will be loaded.

This also enables us to use Next functions for initial props and data fetching. And automatic bundle splitting and such, which is awesome


src/core/components/Header/Login.tsx
```diff
@@ -47,11 +49,17 @@ const HeaderUser = (props: IHeaderUserProps) => (
...
       <div className={style.userMenu} onClick={props.onClick}>
-        <Link to={routes.profile}>Min side: {props.user.profile.preferred_username}</Link>
-        <Link to={routes.home}>Kontakt oss</Link>
-        <Link to={routes.home}>Finn brukere</Link>
-        <Link to={routes.home} onClick={props.logout}>
-          Logg ut
+        <Link {...getMyProfileUrl()}>
+          <a>Min side: {props.user.profile.preferred_username}</a>
+        </Link>
+        <Link href="#">
```
FluidSense, line 55: Har denne aldri gått noensteds? Hahah

FluidSense: Den burde kanskje gå til footern på frontpage, men det får bli et annet PR

src/core/providers/QueryParams.tsx
```diff
@@ -1,16 +1,13 @@
...
 
-export interface IProps extends RouteComponentProps {
+interface IProps {
   children: ReactNode;
 }
 
-export const QueryParamsProvider = ({ location, children }: IProps) => {
-  const queryParams = useQueryParamsState(location);
+export const QueryParamsProvider = ({ children }: IProps) => {
+  const queryParams = useQueryParamsState();
```
FluidSense, line 11: This doesn't need an argument anymore? Neat!

src/core/redux/Store.tsx
```diff
@@ -7,7 +7,8 @@ import { rootReducer } from './reducer';
 
 import { Action, State } from './types';
 
-const reduxDevToolsHook = window.__REDUX_DEVTOOLS_EXTENSION__ ? window.__REDUX_DEVTOOLS_EXTENSION__() : (x: any) => x; // tslint:disable-line no-any
+const reduxDevToolsHook =
+  process.browser && window.__REDUX_DEVTOOLS_EXTENSION__ ? window.__REDUX_DEVTOOLS_EXTENSION__() : (x: any) => x; // tslint:disable-line no-any
 const thunkMiddleware = thunk as ThunkMiddleware<State, Action>;
 
 export const initStore = (initialState?: State) =>
```
FluidSense, line 14: Can't most of this file be squished into a `configureStore`?
https://redux.js.org/redux-toolkit/overview
Default devtools support and thunk middleware! 🎉

src/frontpage/components/Articles/MainArticle.tsx
```diff
@@ -1,23 +1,25 @@
...
-        <div>
-          <h2>{heading}</h2>
-          <Markdown source={ingress.replace(/#[^\s#]/g, (match) => `# ${match.slice(-1)}`)} />
+    <Link {...getArticleUrl(id)}>
+      <a>
+        <div className={style.articleContainer}>
+          <ResponsiveImage image={image} size="sm" type="article" />
+          <div>
+            <h2>{heading}</h2>
+            <Markdown source={ingress.replace(/#[^\s#]/g, (match) => `# ${match.slice(-1)}`)} />
```
FluidSense, line 19: What does this regex-replace do?

src/pages/_app.tsx
```diff
@@ -0,0 +1,45 @@
+import * as Sentry from '@sentry/browser';
+import { Settings as LuxonSettings } from 'luxon';
+import DefaultApp, { AppProps } from 'next/app';
+import React from 'react';
+
+import 'react-day-picker/lib/style.css';
```
FluidSense, line 6: So that's where the react-day-picker css went! Ignore my previous comment then!

src/pages/_app.tsx
```diff
@@ -0,0 +1,45 @@
+import * as Sentry from '@sentry/browser';
```
FluidSense, line 1: What decides whether a page's name should be underscore-prefixed or not?

src/pages/_document.tsx
```diff
@@ -0,0 +1,36 @@
...
+type DocumentProps = ComponentProps<typeof Document>;
+
+const getInitialProps = async (ctx: DocumentContext): Promise<DocumentInitialProps> => {
+  const initialProps = await Document.getInitialProps(ctx);
+  return {
+    ...initialProps,
+  };
+};
+
+const CustomDocument = ({  }: DocumentProps): JSX.Element => {
```
FluidSense, line 13: `({  }: DocumentProps)`
What does this do? An empty bracket?

src/profile/components/MainMenu.tsx
```diff
@@ -1,38 +1,38 @@
...
-import { routes } from '../index';
 import style from './mainMenu.less';
 
-export interface IProps {
-  match: Router.match<{}>;
-}
-
-class MainMenu extends Component<IProps, {}> {
-  public render() {
-    const { path } = this.props.match;
```
FluidSense, line 14: Did you intentionally remove the feature of being able to see what menu element is the current chosen one, is it a mistake or is it being fixed somewhere else?

src/profile/components/Settings/Menu/index.tsx
```diff
@@ -1,21 +1,19 @@
...
+const Menu = () => (
   <div className={style.menu}>
-    <Tab path={routes.main} active={path === routes.main} text="Informasjon" />
-    <Tab path={routes.mail} active={path === routes.mail} text="E-Post" />
-    <Tab path={routes.password} active={path === routes.password} text="Passord" />
-    <Tab path={routes.penalties} active={path === routes.penalties} text="Prikker og suspensjoner" />
-    <Tab path={routes.privacy} active={path === routes.privacy} text="Personvern" />
-    <Tab path={routes.accessCard} active={path === routes.accessCard} text="Adgangskort (NTNU)" />
-    <Tab path={routes.notifications} active={path === routes.notifications} text="Notifikasjoner" />
+    <Tab text="Informasjon" {...appUrls.getProfileSettingsUrl()} />
```
FluidSense, line 10: The same question regarding which is current applies here as well I assume

src/profile/providers/SearchFilter.tsx
```diff
@@ -36,43 +35,21 @@ const INITIAL_STATE: IState = {
...
 
-const parseRange = (rangeString: string): [number, number] | null => {
-  const match = rangeString.match(/^\[\d\,\d\]$/);
-  if (!match) {
-    return null;
-  }
-  const [n, m] = match.map((k) => Number(k));
-  return [n, m];
-};
-
```
FluidSense, line 47: Is this parsing not necessary anymore?

src/profile/providers/SearchFilter.tsx
```diff
@@ -36,43 +35,21 @@ const INITIAL_STATE: IState = {
...
-  const group = queryGroup || undefined;
-  const range = parseRange(queryRange || '') || [1, 6];
-  const setGroup = (newGroup: string) => {
-    if (newGroup.toLowerCase() === 'alle grupper') {
-      setQueryGroup(null);
-    } else {
-      setQueryGroup(newGroup);
-    }
-  };
-  const setRange = (newRange: [number, number]) => setQueryRange(JSON.stringify(newRange));
```
FluidSense, line 74: I have to say this part was confusing as it was previously, but there is no problem in removing it? What did it do?

public/owf.webmanifest
```diff
@@ -0,0 +1,26 @@
+{
```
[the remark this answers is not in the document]
FluidSense: Nvm, I found it in static! I assume it was just moved manually instead of `git mv` then.

src/core/redux/Store.tsx
FluidSense: I mixed up the PR's, this is perhaps resolved in #343?

next-env.d.ts
```diff
@@ -0,0 +1,2 @@
+/// <reference types="next" />
+/// <reference types="next/types/global" />
```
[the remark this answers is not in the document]
oleast: This file is auto generated by Next when a project includes .ts or .tsx files. I'm not 100% sure of why it needs to be done this way, but I'm sure they have a good reason

next.config.js
```diff
@@ -0,0 +1,45 @@
...
+          config.resolve.plugins.push(new TsconfigPathsPlugin());
+        } else {
+          config.resolve.plugins = [new TsconfigPathsPlugin()];
+        }
+        config.module.rules.push({
+          test: /\.(eot|woff|woff2|ttf|svg|png|jpg|jpeg|gif)$/,
+          use: {
+            loader: 'url-loader',
+            options: {
+              limit: 5000,
```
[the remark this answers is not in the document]
oleast: It is a limit on the file size. If I remember my webpack configurations correctly it is used to determine if you inline an image as a base64 string directly in the js code instead of a URL, or if it is copied to the public directory and a URL is placed in the image source.

If the image is small enough is will be much faster and efficient to include the base64 encoded 'data-url', as you don't need a separate request to fetch the image. But if your images is too large it may have a massive impact on the bundle size and initial load times of the app

public/owf.webmanifest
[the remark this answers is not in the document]
oleast: Yeah, just bad copying on my part :sweat_smile: 
Usually works to just copy manually and i will be detected correctly, but not now since i changed the file as well

src/__tests__/sanity.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-ignore
```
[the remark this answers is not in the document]
oleast: There is a problem with 'isolated-modules' [typescript docs](https://www.typescriptlang.org/v2/en/tsconfig#isolatedModules), which is a compiler option that is required with Next, it won't compile without it.

The problem is that files which are not imported anywhere, and do not import any other files will be identified as isolated modules. To this the ts compiler say "no, don't Since this test is just a sanity check it doesn't import anything, it makes things fail. And since we don't have any other test yet, I decided to keep it to have some boilerplate for when we might eventually need it?

src/authentication/components/PrivateLink.tsx
```diff
@@ -1,20 +1,20 @@
...
-import { Link, LinkProps } from 'react-router-dom';
+import Link from 'next/link';
+import React, { Component, ComponentProps } from 'react';
 import { IUserContext, UserContext } from '../providers/UserProvider';
-import style from './link.less';
+// import style from './link.less';
+
+type LinkProps = ComponentProps<typeof Link>;
 
 class PrivateLink extends Component<LinkProps> {
```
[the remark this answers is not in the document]
oleast: I don't think so?

I'm more versed with the functional style of writing typed components, but the first type argument of a component should be the props?
I don't know if the props are determined correctly when just passing a component

src/pages/_document.tsx
oleast: It means nothing, I'm destructuring the props-object, but not actually picking any of props, since they are not used. The thought process was to keep the typed props in the component if anyone were to ever use any of the props. So that the wouldn't have to look up which props a '_document.tsx' component gets

src/authentication/components/PrivateLink.tsx
```diff
@@ -1,20 +1,20 @@
...
   public static contextType = UserContext;
 
   public render() {
     const { user }: IUserContext = this.context;
-    const { className, to, ...props } = this.props;
+    const { href, ...props } = this.props;
```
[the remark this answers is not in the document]
oleast: I can do that :smile:

src/common/components/Forms/DateRangeInput/index.tsx
```diff
@@ -2,7 +2,6 @@ import { DateTime } from 'luxon';
 import React, { FC, useRef } from 'react';
 import { DayPickerProps } from 'react-day-picker';
 import DayPickerInput from 'react-day-picker/DayPickerInput';
-import 'react-day-picker/lib/style.css';
```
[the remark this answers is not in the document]
oleast: All global styles have to be imported in `_app.tsx` with Next, so I moved it there

src/career/containers/DetailView.tsx
```diff
@@ -6,7 +6,7 @@ import React, { Component, ContextType } from 'react';
 import InfoBox from '../components/JobDetails';
 
 export interface IProps {
-  match: { params: { id: string } };
+  opportunityId: number;
```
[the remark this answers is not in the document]
oleast: The id is passed from the page-file that renders the component, which should be `src/pages/career/[opportunityId].tsx`. It now gets the id from the url of the page. The parts in brackets are dynamic parts, and will be passed as variables to the component in that file

src/core/appUrls.ts
```diff
@@ -0,0 +1,31 @@
+import url from 'common/utils/urlHelper';
+
+export const getFrontPageUrl = () => url`/`;
```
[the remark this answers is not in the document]
oleast: We don't need to, no!

Yes, it is a standard feature of the JS language which I think is named `tagged template literals`. It is what styled-components uses, and can be used for many other things, like SQL and just about any other 'string that needs logic behind it'!

Lets say you want to write SQL, but don't want to handle sanitation manually on the variables. You can create a tagged template-literal which will recieve two arguments:
- 1. All the strings, splitted by inserted variables.
- 2. All the variables
E.g: 
``` typescript
const column = '*'
const table = 'users'
sql`SELECT ${column} FROM ${table}`
// gives us:
> ['SELECT ', ' FROM ']
> ['*', 'users']
```
Then we can sanitize the variables and construct a valid SQL statement and return that as a string, ready for the database.
(This is just an example I think would be cool)

src/core/components/Header/Login.tsx
oleast: Har nok aldri gått noe sted nei, vi får fikse det senere :swan:

src/core/providers/QueryParams.tsx
oleast: Managed to go around it, we could have avoided it with React Router as well, but I guess it was never done

src/core/redux/Store.tsx
oleast: Yes, this is fixed in the next PR :P

src/frontpage/components/Articles/MainArticle.tsx
oleast: I think it re-writes headings in markdown:
``` markdown
// from
#heading 
// to
# heading
```

Something with the markdown parsers for OW4 and OWF not handling 'wrong markdown' in the same way. I think OW4 renders it as if it were written correctly (with the space), while OWF handles it as a regular string and not a header.
You'll have to look at #320

henrikhorluck: There is an isskje with the Markdown used on OW4-articles not having the same markdown-standard as our Markdown-library here, so this ensures that:
# Titles appear like this
#Instead of this

src/pages/_app.tsx
oleast: Underscore-prefixed pages are built-in components of Next. Components like `_document` which renders the standard HTML markup (like an `index.html` in Create React App), and `_app` which is rendered as the main file, and wraps around all other pages (useful for including the same logic on all your pages, like a header).

These components have defaults, but we overwrite them when we create a file with the given name.

src/profile/components/MainMenu.tsx
oleast: I Moved the feature down a few levels, into the component which displays the url. Before hooks we had to get the `props.match.path` directly from the child of a `Route` component. With hooks the path is part of a context, which we can consume anywhere we need it in the tree

src/profile/components/Settings/Menu/index.tsx
oleast: I hope the other answer helps here as well :smile:

src/profile/providers/SearchFilter.tsx
oleast: I think I moved these things out of query-parameters and just into local state. Found it unnecessarily complicated to have them as query params at this moment in time

src/profile/providers/SearchFilter.tsx
oleast: We had the state of search filters stored as strings in the url, but i removed it because I found it too complicated to work with. I'm doing this mostly to make things easier to maintain in the long run. Since it used internal React Router things anyways, I decided it was not worth it to re-write it for the Next Router

PR #343, opened 2020-04-08
Add Redux toolkit, implement articles with redux

oleast:
Redux Toolkit is a super simple way of using Redux, which was developed by the Redux team to make it more accessible.

> Purpose
The Redux Toolkit package is intended to be the standard way to write Redux logic. It was originally created to help address three common concerns about Redux:
> - "Configuring a Redux store is too complicated"
> - "I have to add a lot of packages to get Redux to do anything useful"
> - "Redux requires too much boilerplate code"

[Read more about Redux toolkit](https://redux-toolkit.js.org/introduction/quick-start)

This refactors our Redux implementation to Redux toolkit, and gives us some nice features. Articles are also written with toolkit to showcase how it works.


src/articles/slices/articles.ts
```diff
@@ -0,0 +1,137 @@
...
+};
+
+export const articlesSlice = createSlice({
+  name: 'articles',
+  initialState: articlesAdapter.getInitialState(INITIAL_STATE),
+  reducers: {},
+  extraReducers: (builder) => {
+    builder.addCase(fetchArticlesList.pending, (state) => {
+      if (state.loading === 'idle') {
+        state.loading = 'pending';
```
FluidSense, line 65: What is the thought behind this explicit switch? 
If it is to prevent unnecessary rerendering, does Redux have some builtin prevention of that?

src/pages/_app.tsx
```diff
@@ -1,14 +1,19 @@
+import 'core/polyfills';
+
```
FluidSense, line 2: What is this used for here?

src/pages/index.tsx
```diff
@@ -1,9 +1,26 @@
...
 
 export default FrontPage;
+
+FrontPage.getInitialProps = async ({ store }: IContext) => {
+  const state = store.getState();
+  const isFrontPageArticlesPopulated = Boolean(state.articles.frontPageArticleIds.length);
+  const articlesResult = store.dispatch(fetchFrontPageArticles());
+  if (!isFrontPageArticlesPopulated) {
+    await articlesResult;
+  }
```
FluidSense, line 24: Sleek! ✨

src/pages/_app.tsx
oleast: This was a part I forgot about when switching to Next. This imports and defines some globals, for example to make `fetch` work in NodeJS, since that is part of the browser/dom-API and doesn't actually exist by itself in NodeJS

src/articles/slices/articles.ts
oleast: I can agree the the builder might not be the most readable thing, but I think its better than the alternative. 

Regular actions are _super easy_ with redux toolkit, and we'll get to those later when we actually change our state other than just data loading.
Simple example:
``` typescript
const counterSlice = createSlice({
  name: 'counter',
  initialState: 0,
  reducers: {
    increment: (state) => state + 1,
    decrement: (state) => state - 1,
  },
})

const { increment, decrement } = counterSlice.actions;

// Then you can simply use that in a component like so
const IncrementButton = () => {
  const dispatch = useDispatch();
  
  const onClick = () => {
    dispatch(increment())
  };

  return (
    <button onClick={onClick}>Add to counter</button>
  );
};
```

With thunks the story is a little different, and you can look at how it is done here: https://redux-toolkit.js.org/api/createAsyncThunk, which is almost exactly like the builder. This is because the actions are define outside the reducer. The difference is that it doesn't work as well with typescript, so we have to use the builder :slightly_frowning_face: 
The nice thing about the builder is that the return types of the thunk are inferred automatically :smile:

oleast:
> This looks really nice!
> Redux Toolkit might not be the most readable code imo, especially not with the builder, but it's way better than the repetative Redux-coding from before!

I can agree the the builder might not be the most readable thing, but I think its better than the alternative. 

Regular actions are _super easy_ with redux toolkit, and we'll get to those later when we actually change our state other than just data loading.
Simple example:
``` typescript
const counterSlice = createSlice({
  name: 'counter',
  initialState: 0,
  reducers: {
    increment: (state) => state + 1,
    decrement: (state) => state - 1,
  },
})

const { increment, decrement } = counterSlice.actions;

// Then you can simply use that in a component like so
const IncrementButton = () => {
  const dispatch = useDispatch();
  
  const onClick = () => {
    dispatch(increment())
  };

  return (
    <button onClick={onClick}>Add to counter</button>
  );
};
```

With thunks the story is a little different, and you can look at how it is done here: https://redux-toolkit.js.org/api/createAsyncThunk, which is almost exactly like the builder. This is because the actions are define outside the reducer. The difference is that it doesn't work as well with typescript, so we have to use the builder :slightly_frowning_face: 
The nice thing about the builder is that the return types of the thunk are inferred automatically :smile:

src/pages/index.tsx
oleast: :100:

src/articles/slices/articles.ts
FluidSense: I see! Do you know if it will rerender every time if it is not included?

FluidSense:
@oleast That very link on how smooth it _could be_ and also the argument of why it is not written that way with TypeScript was the source of my comment 😄 
Overall, Redux Toolkit is definitely an improvement!

src/articles/slices/articles.ts
oleast: It seems i picked it from a more customized version of their examples which handled 'only requesting one at a time' in a specific way which we are not. I'll make this more general :smile:

oleast:
> @oleast That very link on how smooth it _could be_ and also the argument of why it is not written that way with TypeScript was the source of my comment 
> Overall, Redux Toolkit is definitely an improvement!

I see, got a bit carried away there :sweat_smile: 
I guess I'm just exited about removing some of the enormous amount of boilerplate that Redux with typescript used to require :package:

src/articles/slices/articles.ts
FluidSense: Neat!

oleast:
@FluidSense 
What do you think in general about implementing this as out standard way of handling data-fetching and state management?
If you think it works, could we be ready to merge?

PR #344, opened 2020-04-15
Switch frontpage offlines to ResponsiveImage

oleast:
^Title

### DIFF
--- a/src/frontpage/components/Offline/OfflineCarousel.tsx
+++ b/src/frontpage/components/Offline/OfflineCarousel.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import ResponsiveImage from 'common/components/ResponsiveImage';
 import { IRefObject } from 'common/hooks/useRefMap';
 import { IOfflineIssue } from 'frontpage/models/Offline';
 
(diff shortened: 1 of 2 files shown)


FluidSense:
Could it be an idea to add the img lazy tag to carousel as well in this PR, to shorten preload?

oleast:
> Could it be an idea to add the img lazy tag to carousel as well in this PR, to shorten preload?

I think we should look into it, but I would like to have offlines rendered with SSR on the front-page before doing so, to be able to check how the actual loading will be :smile: 

It could potentially be a thing we implement for _all_ responsive images if it works well enough, so we can just include it in that compoent?

FluidSense:
So changing them to ResponsiveImages is what is needed for SSR offlines? Cool stuff; get it merged!

oleast:
Changing to ResponsiveImage is not make the images actually work. We made a change in the back-end which automatically makes a ResponsiveImage when an offline is uploaded, and disabled the old way. All Offlines uploaded after that had a missing image.

I ment that it would be nice to have them loaded with SSR before we test different loading-systems for the images. Since the images will be part of the initial HTML we can see how the lazy loading will change the way they are loaded

PR #346, opened 2020-04-18
Refactor career pages to Redux

### LINKED ISSUE #257: Sort opportunities with featured appearing first
Career opportunities should be sorted to make featured ones appear on top.
They have the `featured` flag set in the model.

**Describe alternatives you've considered**
Featured opportunities should probably also add some context to why they are sorted to the top.
We could do this by adding an icon and some text to the card.
Something like a flag or a star below the image. Golden/yellow/orange text saying 'Fremhevet jobbmulighet' or something.

**Additional context**
They have the `featured` flag set in the model.

oleast:
Depends on https://github.com/dotkom/onlineweb4/pull/2463

- Move career state from context to Redux
- Handle searching by calling the API instead of in the front-end
- Add Graph meta tags to career pages
- Re-enable SSR data-fetching for career detail page with getInitialProps

Resolves #257 in the process, because sort comparing is needed for the search to work correctly anyways


src/career/api/index.ts
```diff
@@ -1,21 +1,33 @@
...
+  end__lte?: string;
+  deadline__gte?: string;
+  deadline__lte?: string;
+  company?: number;
+  company__in?: number | number[];
+  employment?: number;
+  employment__in?: number | number[];
+}
+
+export const listCareerOpportunities = listResource<ICareerOpportunity, IFilters>(API_URL);
```
FluidSense, line 27: Do I understand it correctly that listResource and retrieveResource is part of a generic CRUD-set of operations?
Which is then used as part of thunk actions?

src/core/redux/Store.tsx
```diff
@@ -9,6 +13,10 @@ export const initStore = (initialState: {} = {}) => {
     preloadedState: initialState,
     reducer: {
       articles: articlesReducer,
+      careerLocations: careerLocationsReducer,
+      careerOpportunities: careerOpportunitiesReducer,
+      companies: companiesReducer,
+      jobTypes: jobTypesReducer,
```
FluidSense, line 19: What is the benefit/reason of 4 reducers instead of 1 career reducer? Are they being used differently / outside of career opportunities?

src/pages/career/[opportunityId].tsx
```diff
@@ -1,19 +1,35 @@
...
-  const opportunityId = Number(router.query.opportunityId);
+import { careerOpportunitySelectors, fetchCareerOpportunityById } from 'career/slices/careerOpportunities';
+import { Store } from 'core/redux/Store';
+
+interface IContext extends NextPageContext {
+  store: Store;
+}
+
+interface IProps {
+  opportunityId: number;
```
Comment (FluidSense) [inline src/pages/career/[opportunityId].tsx:14]: Just to see if I remember correctly: This is gotten from Next router, right? Based on the id given in the url, which will then match a file with bracketed name [opportunityId].tsx?

src/pages/career/[opportunityId].tsx
```diff
@@ -1,19 +1,35 @@
...
-      </CareerOpportunities>
+      <DetailView opportunityId={opportunityId} />
     </section>
   );
 };
 
+CareerDetailPage.getInitialProps = async ({ query, store }: IContext) => {
+  const opportunityId = Number(query.opportunityId);
+  const isOpportunityInStore = Boolean(careerOpportunitySelectors.selectById(store.getState(), opportunityId));
+  const result = store.dispatch(fetchCareerOpportunityById(opportunityId));
```
Comment (FluidSense) [inline src/pages/career/[opportunityId].tsx:28]: Is there a difference in getting store as a prop and using `store.dispatch` vs. importing `useDispatch()`?

FluidSense:
Cool! This also fixes the 404-flash upon entering "Karriere", right?
Also, preview of the filters will not function correctly without https://github.com/dotkom/onlineweb4/pull/2463, correct?

src/pages/career/[opportunityId].tsx
Comment (oleast) [reply src/pages/career/[opportunityId].tsx:28]: There is a difference here, since `Page.getInitialProps` is part of the Next runtime, not React itself.
`useDispatch` will get the dispatch function from the current store through the redux context from the redux provider.
Since `getInitialProps` is run outside of the React component tree there is no provider/context to the the current store/dispatch function from.

To allow our `getInitialProps` function to access the redux store and dispatch functions the store is passed to all `getInitialProps` functions from the `_app` component, which is where the store is created (or retrieved if it already exists). This is done under the hood by `next-redux-wrapper`

src/pages/career/[opportunityId].tsx
Comment (oleast) [reply src/pages/career/[opportunityId].tsx:14]: Yes, exactly. Since this the name of this file contains a bracketed variable name, that part of the url will become a variable in the query object :smile:

src/core/redux/Store.tsx
oleast: The idea is keep all kinds of entities separate from each other. There is no absolute or correct way to handle this though.

My thoughts are that each kind of entity should be separated into each own reducer because they are re-used in other parts of the app or duplicated in this part.
E.g. each career opportunity is connected to a company, but a single company  may have many career opportunities. This would lead us to have duplicates of that company on our store.
If we put the companies in their own reducer we get to use the store as a bit of a cache, since when you go to the company details page from an opportunity page the company data is already populated in the store, and you won't have to wait for it to be fetched to view the page.

This could also let us change how data is fetached more easily. We could change the career api to not deliver company data at initial load.

src/career/api/index.ts
oleast: Yes, they are an abstraction I made which maps directly to Django Rest Framework operations. It does a lot of things for you, and is supposed to make typings work very strictly for all operations, but I'm not really happy with it.

I might hope to clean it up some day, but it actually works well, so I haven't thought too much about it

oleast:
> Cool! This also fixes the 404-flash upon entering "Karriere", right?

Yes it does :smile: 

> Also, preview of the filters will not function correctly without [dotkom/onlineweb4#2463](https://github.com/dotkom/onlineweb4/pull/2463), correct?

Correct

src/pages/career/[opportunityId].tsx
Comment (FluidSense) [reply src/pages/career/[opportunityId].tsx:28]: I see! Neat stuff!

src/pages/career/[opportunityId].tsx
```diff
@@ -1,19 +1,35 @@
...
 
-const CareerDetailPage = () => {
-  const router = useRouter();
-  const opportunityId = Number(router.query.opportunityId);
+import { careerOpportunitySelectors, fetchCareerOpportunityById } from 'career/slices/careerOpportunities';
+import { Store } from 'core/redux/Store';
+
+interface IContext extends NextPageContext {
+  store: Store;
+}
```
Comment (FluidSense) [inline src/pages/career/[opportunityId].tsx:11]: Will all pages with getInitialProps have this as interface?
If yes, could this be generalized somewhere and exported?

Comment (oleast) [reply src/pages/career/[opportunityId].tsx:11]: Yes, all pages will have it.
I agree, should be generalized, just didn't think to do it yet :P

PR #348, opened 2020-04-18
Refactor events to Redux

### LINKED ISSUE #281: Event detail does not fetch event data
**Describe the bug**
If you click an event when not logged in everything is fine. But when you try to log in while already on the event page no event info is rendered.

**To Reproduce**
1. Go to beta.online.ntnu.no
2. Click on an event
3. Click on login
4. ???
5. profit

**Expected behavior**
The event should show up correctly as it does when you refresh the page.

**Screenshots**
bug

 - OS: Linux 4.19.66-1-MANJARO
 - Browser: Chrome 
 - Version: 76

FluidSense:
Resolves #281


src/events/slices/events.ts
```diff
@@ -0,0 +1,118 @@
...
+      builder.addCase(fetchEventsByMonth.fulfilled, (state, action) => {
+        state.loading = 'idle';
+        eventsAdapter.addMany(state, action.payload);
+      }),
+      builder.addCase(fetchEventsByMonth.rejected, (state, action) => {
+        state.loading = 'idle';
+        state.error = action.error;
+      });
+  },
+});
```
FluidSense, line 116: This really feels repetative! 😅 
I wonder if it's possible to merge their handling, but keep the action names?

oleast: I have an idea that might work:
Move the api abstraction to from the API functions to thunks:
We have functions like `getListEvents`, `getImageEvents` and `getCalendarEvents` that all are an abstraction of the generic `getEvents`.

If we instead have a generic `fetchEvents` thunk which calls `getEvents`. This thunk can be used in the reducer and add multiple events like the current `fetchEvents`.
If we then have the `fetchEventsByMonth`, `fetchImageEvents` and `fetchEventList` be an abstraction of that like so:

``` typescript
// Just an example, don't know if it works
export const fetchEventList = createAsyncThunk('events/fetchList', async (_, { dispatch ) => {
  const response = await (dispatch(fetchEvents({
    event_end__gte: DateTime.local().toISODate(),
    page_size: 10,
  }));
  return unwrapResult(response);
});
```

That way we have named thunks which do pre-defined things, but only a single thunk we can abstract to our reducer

src/core/redux/Store.tsx
```diff
@@ -19,6 +20,7 @@ export const initStore = (initialState: {} = {}) => {
       jobTypes: jobTypesReducer,
       payments: paymentsReducer,
       shop: shopReducer,
+      events: eventsReducer,
```
oleast, line 23: Don't know if its worth keeping to do it, but i tried to sort the reducer names alphabetically

src/events/components/ImageView/index.tsx
```diff
@@ -46,14 +46,13 @@ const isPopulated = (imageEvents: IState) => {
 };
 
 export const ImageView = ({  }: IProps) => {
-  const { eventList, updateEventList } = useContext(EventsRepo);
+  const dispatch = useDispatch();
+  const eventList = useSelector((state) => eventSelectors.selectAll(state));
+  // const { eventList, updateEventList } = useContext(EventsRepo);
```
oleast, line 51: Could remove this comment, won't need it ever again :hear_no_evil:

src/events/components/ImageView/index.tsx
```diff
@@ -46,14 +46,13 @@ const isPopulated = (imageEvents: IState) => {
...
 
   /** Fetch events to repo on mount */
```
oleast, line 53: This comment may not be strictly correct anymore?

src/events/components/DetailView/index.tsx
```diff
@@ -16,8 +18,21 @@ export interface IProps {
 }
 
 export const DetailView = ({ eventId }: IProps) => {
-  const { eventMap } = useContext(EventsRepo);
-  const event = eventMap.get(eventId) || mockEvent;
+  const dispatch = useDispatch();
+  const event = useSelector((state) => eventSelectors.selectById(state, eventId));
+  const isPending = useSelector((state) => state.articles.loading === 'pending');
```
oleast, line 23: This uses the loading state of articles, not events

src/pages/events/[eventId].tsx
```diff
@@ -1,12 +1,29 @@
...
+import { eventSelectors, fetchEventById } from 'events/slices/events';
+import { NextPageContext } from 'next';
+
+interface IContext extends NextPageContext {
+  store: Store;
+}
 
 const EventDetailPage = () => {
   const router = useRouter();
   const eventId = Number(router.query.eventId);
```
Comment (oleast) [inline src/pages/events/[eventId].tsx:15]: Since `getInitialProps` has to return _something_ I returned the `id` in the other examples.
Anything returned from `getInitialProps` will be passed as props to the component, which means we can just use the `eventId` from props instead of getting it from the router again :smile:

src/core/redux/Store.tsx
henrikhorluck: Wouldn't it be easier to sort this through linting, like https://eslint.org/docs/rules/sort-keys, og https://palantir.github.io/tslint/rules/object-literal-sort-keys/?

oleast: I don't think we should turn it on for the whole project, but this object will be huge, and I think it is beneficial just here.

Is it possible to selectively turn linting rules on for specific parts of code?

src/events/components/ImageView/index.tsx
FluidSense: True, it's store and not repo now 😄

src/events/components/ImageView/index.tsx
FluidSense: Oops! 😅

src/events/slices/events.ts
FluidSense: This works like a charm! The actions becomes a bit harder to read in the devtools, as there is a lot of fetch actions, but this makes way more sense to me than to repeat the reducer methods for every single of fetch request!

src/core/redux/Store.tsx
henrikhorluck: Yes you can https://palantir.github.io/tslint/usage/rule-flags/
syntax:
```js
/* tslint:enable:rule1 rule2 */
{ code... }
/* teslint:disable: rule1 rule2 */
```
I don't think there is a way to reset it back to default value from config

src/events/slices/events.ts
```diff
@@ -0,0 +1,116 @@
...
+  const left = getTypeEvents([EventTypeEnum.BEDPRES]);
+  const middle = getTypeEvents([EventTypeEnum.KURS]);
+  const right = getTypeEvents([
+    EventTypeEnum.SOSIALT,
+    EventTypeEnum.UTFLUKT,
+    EventTypeEnum.EKSKURSJON,
+    EventTypeEnum.ANNET,
+  ]);
+  const responses = await Promise.all([left, right, middle]);
+  responses.forEach((payloadCallback) => unwrapResult(payloadCallback));
```
oleast, line 55: Will this actually return the responses?
As I understand `forEach` it doesn't actually change the array?

src/events/slices/events.ts
```diff
@@ -0,0 +1,116 @@
...
+export const eventSelectors = eventsAdapter.getSelectors<State>((state) => state.events);
+
+export const fetchEvents = createAsyncThunk('events/fetchMultiple', async (options?: IEventAPIParameters) => {
+  const events = await getEvents(options);
+  return events;
+});
+
+export const fetchEventById = createAsyncThunk('events/fetchById', async (eventId: number) => {
+  const events = await getEvent(eventId);
+  return events;
```
oleast, line 22: `events` -> `event`?

PR #349, opened 2020-04-19
Move static assets from OW4 to OWF

oleast:
We had a lot of images that relied directly on OW4, I moved them here instead


oleast:
It has the effect you noted, as well as being one less thing we have to change when the assets are removed from OW4. In the future it would be nice to have it as a pure API

PR #350, opened 2020-04-19
Migrate from TSLint to ESLint

FluidSense:
## Why?
TSLint has been the recommended linter in the past but now TSLint is deprecated and ESLint is taking over its duties.


.eslintrc.js
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "globals": {
+	"React": "writable"
+  },
+    "settings" : {
+        "react": {
+            "version": "detect"
+        }
+    }
+};
```
oleast, line 24: We have a quite similar setup for most of our other repos, maybe just copy that? https://github.com/dotkom/glowing-fortnight/blob/master/.eslintrc.js

package.json
```diff
@@ -12,11 +12,11 @@
...
     "lint": "yarn lint-ts ; yarn lint-less ; yarn lint-prettier",
-    "lint-ts": "tslint -c tslint.json 'src/**/*.{ts,tsx}'",
+    "lint-ts": "eslint -c eslint.js 'src/**/*.{ts,tsx}'",
     "lint-less": "stylelint src/**/*.less",
     "lint-prettier": "prettier-check 'src/**/*.{ts,tsx,less}'",
     "lint:fix": "yarn lint-ts:fix && yarn lint-less:fix && yarn lint-prettier:fix",
-    "lint-ts:fix": "tslint --fix -c tslint.json 'src/**/*.{ts,tsx}'",
+    "lint-ts:fix": "eslint --fix -c eslint.js 'src/**/*.{ts,tsx}'",
     "lint-less:fix": "yarn lint-less --fix",
     "lint-prettier:fix": "prettier-check --write 'src/**/*.{ts,tsx,less}'",
```
oleast, line 21: Could be have the same basic command structure as our other repos?
https://github.com/dotkom/glowing-fortnight/blob/master/package.json#L12-L15

FluidSense: I agree! I'll make it so!

FluidSense: Hm, I'm a bit uncertain on the commands. I see we're using another linter to lint `.less`-files.
I could include it in the basic command and run both linters in one command, since linting is one job in my mind.
However, one could perhaps want to know exactly which linter reports about which errors and thus run them separately.
What do you think, @oleast ?

src/common/resources/utils.ts
```diff
@@ -22,16 +22,12 @@ export const handleReadOnlyResponse = async <InputData, OutputData>(
...
   if (response.ok) {
     if (response.status === 204) {
       return {
         status: 'success',
-        // Handle no-content responses as null.
-        // This is hind of dangerous, since null has to be defined as OutputData
-        // When the resouce is created.
-        // tslint:disable-next-line no-any
-        data: (null as any) as OutputData,
+        data: null,
```
oleast, line 30: The idea here is that we are enforcing that the response can't actually be null, which means we won't have to handle the null case all the time. Allowing the return type here to be null may remove that feature

FluidSense: I see. As I read it, we are currently enforcing that the response is _always null_ on 204, but letting TypeScript treat `null` as the type parameter `OutputData`, is that what you meant?

This can result in very quirky inconsistencies if I pass `number` as the type parameter and expects a 200, but got a 204 because of a misconfiguration of the API-point. TypeScript will allow me to use the result in calculations, for example 
```js
const user = handleResponse<User>(fetchUser());
const balance = user.balance + 100;
```
User is actually null here, since fetchUser resulted in 204 (perhaps since I used the wrong endpoint when making the fetchUser()-function. 😅 )
`balance` will then be `NaN` since what I actually did was `null.balance + 100` and `null.balance` is `undefined`.

I guess this is why the warning was there in the comment, but that would be a warning shown only those peeking into the method would see.
I suggest handling `null` one level higher, in the 6 methods `listResource`, `createResource`, etc.
Those methods know what type of data they want, and can throw an error accordingly.
What do you think about that way of doing it?

FluidSense: I've included suggestion of implementation the way I mentioned in 2c6782e.
If that still feels wrong, I'll revert the change and use `unknown` instead of `any` to satisfy TypeScript.

.eslintrc.js
FluidSense: Copied!

package.json
oleast: I like the idea of running one command to check the entire thing, but for things like CI we should be able to run the all separately.
I think I like the setup you have here :smile: 

On another note, do we actually need the `prettier-check` dependency, it seems from the other repos that prettier can do that fine by itself?

src/common/resources/utils.ts
oleast: The problem I see is that we are adding a possible `null` value to all requests made with the functions, which because of typescript you have to handle every time you use it. I think that is a hard price to pay for an edge case of API-endpoints that are misusing HTTP 204 (no content)?

FluidSense: I definitely agree to that being a hard price to pay!
Are there requests which will not expand upon the configured CRUD-requests, but still use handleResponse? I like the idea that every request can use the CRUD-base requests, and let those base requests handle null!

package.json
FluidSense: Good spot! Removed `prettier-check`!

src/common/resources/utils.ts
oleast: There are no other requests then these who user `handleResponse`, but I see that these CRUD operations are missing custom actions, and other requests, which means you can't use them for everything.

All in all these _resource_ functions are a lot of magic, which may be nice, but maybe a bit much? And as I said they can't do everything.

On the other hand I think the other API utils we user other places has a very clunky usage api, their function signatures are a bit strange.

I would hope we could combine them in one way or another, to make _one_ way to do requests.

Now that we actually get error handling with Redux Thunks catching thrown error I think we could just let our API throw on 4-5xx responses, instead or returning that complex object we do now.

PR #354, opened 2020-04-19
Captcha-UI on attendance events

FluidSense:


src/events/components/DetailView/AttendanceEvent.tsx
```diff
@@ -55,6 +56,9 @@ const AttendanceEvent = ({ event }: IAttendanceEventProps) => {
   const registrationStart = DateTime.fromISO(event.registration_start);
   const registrationEnd = DateTime.fromISO(event.registration_end);
   const cancellationDeadline = DateTime.fromISO(event.unattend_deadline);
+  const sitekey: string = process.env.RECAPTCHA_PUBLIC_KEY || '';
```
oleast, line 59: This should be defined in the `/src/common/constants` folder like everything else we get from env

src/events/components/DetailView/AttendanceEvent.tsx
```diff
@@ -55,6 +57,8 @@ const AttendanceEvent = ({ event }: IAttendanceEventProps) => {
   const registrationStart = DateTime.fromISO(event.registration_start);
   const registrationEnd = DateTime.fromISO(event.registration_end);
   const cancellationDeadline = DateTime.fromISO(event.unattend_deadline);
+  // @ts-ignore
```
henrikhorluck, line 60: Why is this ignored?

oleast: Because it is not used yet, but it was left as something fore the others to implement

FluidSense: Yes, exactly what @oleast said!
I should perhaps have included a comment about that with a TODO: Remove when used.

Issue #45, opened 2018-10-26
Add carousel or similar for large events if several images


plusk:
Events like ITEX don't have a dedicated image, but do have several companies tied to the event. What I added in #43 only pulls out the first of the images. 

What needs to be done
- Get all the images, not just the first
- Add a carousel, or similar, and put the images in it

In OW4, the merging/getting is done [here](https://github.com/dotkom/onlineweb4/blob/develop/assets/frontpage/events/containers/EventsContainer.js#L6), while the carousel is [here](https://github.com/dotkom/onlineweb4/blob/develop/assets/frontpage/events/components/EventImage.js). Note that the carousel comes from React-bootstrap, and we'd have to make our own.

PR #356, opened 2020-04-21
Refactor events to new API

### LINKED ISSUE #167: Undefined company event for 'kurs' and 'bedpres'
Expecting at least a single company event, but the list is empty.

https://sentry.io/dotkom/onlineweb-frontend/issues/853657848/?query=is%3Aunresolved

### LINKED ISSUE #353: Display attendance possibility
Display whether or not you ackchyually can attend the event, and display the login button if you can

oleast:
Resolves #353 (somewhere in there we have a selector and a ts-ignore)
Resolves #167 (by selecting images as part of the api)

Obviously relies on https://github.com/dotkom/onlineweb4/pull/2468 to work

### DIFF
--- a/src/authentication/models/User.ts
+++ b/src/authentication/models/User.ts
@@ -1,4 +1,4 @@
-import { User, Profile } from 'oidc-client';
+import { Profile, User } from 'oidc-client';
 
 export interface IAuthUser extends User {
   profile: IAuthProfile;
@@ -11,3 +11,10 @@ export interface IAuthUser extends User {
 export type IAuthProfile = NonNullable<Profile>;
 
 export const isOnlineUser = (user: User | IAuthUser): user is IAuthUser => user.hasOwnProperty('preferred_username');
+
+export interface IUserName {
+  id: number;
+  first_name: string;
+  last_name: string;
+  username: string;
+}
--- a/src/companies/slices/companies.ts
+++ b/src/companies/slices/companies.ts
@@ -1,5 +1,6 @@
-import { createEntityAdapter, createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { createAsyncThunk, createEntityAdapter, createSlice, PayloadAction, SerializedError } from '@reduxjs/toolkit';
 
+import { retrieveCompany } from 'companies/api';
 import { ICompany } from 'core/models/Company';
 import { State } from 'core/redux/Store';
 
@@ -12,11 +13,24 @@ const companiesAdapter = createEntityAdapter<ICompany>({
 
 export const companySelectors = companiesAdapter.getSelectors<State>((state) => state.companies);
 
+export const fetchCompanyById = createAsyncThunk('companies/fetchById', async (companyId: number) => {
+  const response = await retrieveCompany(companyId);
+  if (response.status === 'success') {
+    return response.data;
+  } else {
+    throw response.errors;
+  }
+});
+
 interface IState {
-  entities: Record<string, ICompany>;
+  loading: 'idle' | 'pending';
+  error: SerializedError | null;
+  entities: Record<number, ICompany>;
 }
 
 const INITIAL_STATE: IState = {
+  loading: 'idle',
+  error: null,
   entities: {},
 };
 
@@ -32,6 +46,19 @@ export const companiesSlice = createSlice({
       const companies = action.payload;
       companiesAdapter.addMany(state, companies);
     },
+  },
+  extraReducers: (builder) => {
+    builder.addCase(fetchCompanyById.pending, (state) => {
+      state.loading = 'pending';
+    });
+    builder.addCase(fetchCompanyById.fulfilled, (state, action) => {
+      state.loading = 'idle';
+      companiesAdapter.addOne(state, action.payload);
+    });
+    builder.addCase(fetchCompanyById.rejected, (state, action) => {
+      state.loading = 'idle';
+      state.error = action.error;
+    });
   },
 });
 
--- a/src/core/redux/Store.tsx
+++ b/src/core/redux/Store.tsx
@@ -5,8 +5,13 @@ import { careerLocationsReducer } from 'career/slices/careerLocations';
 import { careerOpportunitiesReducer } from 'career/slices/careerOpportunities';
 import { jobTypesReducer } from 'career/slices/jobTypes';
 import { companiesReducer } from 'companies/slices/companies';
+import { attendanceEventsReducer } from 'events/slices/attendanceEvents';
+import { attendeesReducer } from 'events/slices/attendees';
 import { eventsReducer } from 'events/slices/events';
-import { paymentsReducer } from 'payments/reducers';
+import { ruleBundlesReducer } from 'events/slices/ruleBundles';
+import { onlineGroupsReducer } from 'groups/slices/onlineGroups';
+import { transactionsReducer } from 'payments/reducers/transactions';
+import { paymentsReducer } from 'payments/slices/payments';
 import { shopReducer } from 'shop/reducers';
 
 export const initStore = (initialState: {} = {}) => {
@@ -15,13 +20,18 @@ export const initStore = (initialState: {} = {}) => {
     /* eslint sort-keys: "error" */
     reducer: {
       articles: articlesReducer,
+      attendanceEvents: attendanceEventsReducer,
+      attendees: attendeesReducer,
       careerLocations: careerLocationsReducer,
       careerOpportunities: careerOpportunitiesReducer,
       companies: companiesReducer,
       events: eventsReducer,
       jobTypes: jobTypesReducer,
+      onlineGroups: onlineGroupsReducer,
       payments: paymentsReducer,
+      ruleBundles: ruleBundlesReducer,
       shop: shopReducer,
+      transactions: transactionsReducer,
     },
     /* eeslint sort-keys: "off" */
   });
--- a/src/core/redux/Store.tsx
+++ b/src/core/redux/Store.tsx
@@ -33,7 +33,7 @@ export const initStore = (initialState: {} = {}) => {
       shop: shopReducer,
       transactions: transactionsReducer,
     },
-    /* eeslint sort-keys: "off" */
+    /* eslint sort-keys: "off" */
   });
 };
 
--- a/src/events/api/ruleBundles.ts
+++ b/src/events/api/ruleBundles.ts
@@ -1,0 +1,12 @@
+import { listResource, retrieveResource } from 'common/resources';
+import { IQueryObject } from 'common/utils/queryString';
+import { IRuleBundle } from 'events/models/RuleBundles';
+
+const RULE_BUNDLE_API_URL = '/api/v1/event/rule-bundles';
+
+interface IRuleBundleFilters extends IQueryObject {
+  event?: number;
+}
+
+export const listRuleBundles = listResource<IRuleBundle, IRuleBundleFilters>(RULE_BUNDLE_API_URL);
+export const retrieveRuleBundle = retrieveResource<IRuleBundle>(RULE_BUNDLE_API_URL);
--- a/src/events/components/CalendarView/CalendarHoverCard.tsx
+++ b/src/events/components/CalendarView/CalendarHoverCard.tsx
@@ -3,19 +3,30 @@ import { faCalendarAlt as CalendarAlt } from '@fortawesome/free-regular-svg-icon
 import { faCalendarCheck as CalendarCheck } from '@fortawesome/free-regular-svg-icons/faCalendarCheck';
 import { faMapMarkedAlt as MapMarked } from '@fortawesome/free-solid-svg-icons/faMapMarkedAlt';
 import { FontAwesomeIcon as Icon } from '@fortawesome/react-fontawesome';
-import { IEvent } from 'events/models/Event';
 import { DateTime } from 'luxon';
-import React from 'react';
+import React, { FC } from 'react';
+
+import { useSelector } from 'core/redux/hooks';
+import { State } from 'core/redux/Store';
+import { IEvent } from 'events/models/Event';
+import { attendanceEventSelectors } from 'events/slices/attendanceEvents';
+import { eventSelectors } from 'events/slices/events';
+
 import EventImage from '../EventImage';
 import style from './calendar.less';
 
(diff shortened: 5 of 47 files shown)


oleast:
There are a few small regressions, alt least one which has to be fixed before merging:

- Doesn't display amount of registered attendees on the front-page anymore. (We'll have to fix this...)
- The canonical event title is used everywhere now. Maybe this works just as well? Some places it just displayed the company name, but that logic was just so unpredictable.
- Woops, I don't think rule-bundles are actually loaded
- Event search doesn't work with company names anymore. But we'll get to that later

src/events/components/DetailView/AttendanceEvent.tsx
```diff
@@ -1,64 +1,37 @@
...
-    </Block>
-  );
-};
-
-const AttendanceEvent = ({ event }: IAttendanceEventProps) => {
-  const registrationStart = DateTime.fromISO(event.registration_start);
-  const registrationEnd = DateTime.fromISO(event.registration_end);
-  const cancellationDeadline = DateTime.fromISO(event.unattend_deadline);
+const AttendanceEvent: FC<IProps> = ({ eventId }) => {
   // @ts-ignore
```
henrikhorluck, line 22:
```suggestion
  // tslint:disable-next-line: no-unused-variable
```

src/core/redux/Store.tsx
```diff
@@ -5,22 +5,32 @@ import { careerLocationsReducer } from 'career/slices/careerLocations';
...
+import { ruleBundlesReducer } from 'events/slices/ruleBundles';
+import { onlineGroupsReducer } from 'groups/slices/onlineGroups';
+import { transactionsReducer } from 'payments/reducers/transactions';
+import { paymentsReducer } from 'payments/slices/payments';
 import { shopReducer } from 'shop/reducers';
 
 export const initStore = (initialState: {} = {}) => {
   return configureStore({
     preloadedState: initialState,
     reducer: {
```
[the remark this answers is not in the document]
oleast: I think we'll take these things with the incoming eslint PR

src/authentication/models/User.ts
```diff
@@ -11,3 +11,10 @@ export interface IAuthUser extends User {
 export type IAuthProfile = NonNullable<Profile>;
 
 export const isOnlineUser = (user: User | IAuthUser): user is IAuthUser => user.hasOwnProperty('preferred_username');
+
+export interface IUserName {
+  id: number;
+  first_name: string;
+  last_name: string;
+  username: string;
+}
```
FluidSense, line 20: Is this a subset of User or IAuthUser?
If so, perhaps a [Pick<T, K> type](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktk) could be used?
`export const IUserName = Pick<IAuthUser, 'id' | 'first_name' | 'last_name' | 'username'>;`

Or is it a special type?

src/companies/slices/companies.ts
```diff
@@ -12,11 +13,24 @@ const companiesAdapter = createEntityAdapter<ICompany>({
...
+    return response.data;
+  } else {
+    throw response.errors;
+  }
+});
+
 interface IState {
-  entities: Record<string, ICompany>;
+  loading: 'idle' | 'pending';
+  error: SerializedError | null;
```
FluidSense, line 27: Neat typing!

src/core/redux/Store.tsx
```diff
@@ -15,13 +20,18 @@ export const initStore = (initialState: {} = {}) => {
...
       companies: companiesReducer,
       events: eventsReducer,
       jobTypes: jobTypesReducer,
+      onlineGroups: onlineGroupsReducer,
       payments: paymentsReducer,
+      ruleBundles: ruleBundlesReducer,
       shop: shopReducer,
+      transactions: transactionsReducer,
     },
     /* eeslint sort-keys: "off" */
```
FluidSense, line 36:
```suggestion
    /* eslint sort-keys: "off" */
```
I made a typo, can we fix it here? 😅

src/events/api/ruleBundles.ts
```diff
@@ -0,0 +1,12 @@
+import { listResource, retrieveResource } from 'common/resources';
+import { IQueryObject } from 'common/utils/queryString';
+import { IRuleBundle } from 'events/models/RuleBundles';
+
+const RULE_BUNDLE_API_URL = '/api/v1/event/rule-bundles';
+
+interface IRuleBundleFilters extends IQueryObject {
+  event?: number;
```
FluidSense, line 8: Is this the primaryKey for an event?
Can I suggest to change it to `eventId` if so?

src/events/components/CalendarView/CalendarHoverCard.tsx
```diff
@@ -3,19 +3,30 @@ import { faCalendarAlt as CalendarAlt } from '@fortawesome/free-regular-svg-icon
...
-const CalendarHoverCard = ({ location, image, company_event, attendance_event, title, event_start }: IEvent) => {
-  const registrationStart =
-    attendance_event && DateTime.fromISO(attendance_event.registration_start).toFormat('d MMM HH:mm');
-  const eventStart = DateTime.fromISO(event_start).toFormat('d MMM HH:mm');
+interface IProps {
+  eventId: number;
+}
+
+const CalendarHoverCard: FC<IProps> = ({ eventId }) => {
+  const event = useSelector((state) => eventSelectors.selectById(state, eventId) as IEvent);
```
FluidSense, line 23: `..eventSelectors.selectById(state, eventId) as IEvent`
Do you have to type cast it to event? Is it not inferred from the adapters' type?

src/events/components/CalendarView/CalendarTile.tsx
```diff
@@ -55,9 +55,9 @@ export const CalendarFillerTiles = ({ days }: { days: number[] }) => (
 export const CalendarEvent = (event: IEvent) => (
   <Link {...getEventUrl(event.id)}>
     <a>
-      <HoverCard card={<CalendarHoverCard {...event} />}>
+      <HoverCard card={<CalendarHoverCard eventId={event.id} />}>
         <p className={style.title} style={{ background: getEventColor(event.event_type) }}>
-          {isCompanyEvent(event.event_type, event.company_event) ? event.company_event[0].company.name : event.title}
```
FluidSense, line 60: ..Did we use to show the company name instead of the event title?

src/events/components/DetailView/index.tsx
```diff
@@ -23,15 +24,17 @@ export const DetailView = ({ eventId }: IProps) => {
   const isPending = useSelector((state) => state.events.loading === 'pending');
 
   useEffect(() => {
+    window.scrollTo(0, 0);
```
FluidSense, line 27: Why is scrolling to the top needed?

src/events/components/EventImage.tsx
```diff
@@ -1,18 +1,17 @@
...
-  companyEvents: ICompanyEvent[];
+  images: IResponsiveImage[];
   size: keyof IResponsiveImageSizes;
   color?: string;
 }
 
-const EventImage: FC<IProps> = ({ image, companyEvents: [companyEvent], color, ...props }) => {
-  const img = companyEvent ? companyEvent.company.image : image;
+const EventImage: FC<IProps> = ({ images, color, ...props }) => {
+  // TODO: implement a way to display multiple event images.
```
FluidSense, line 13: Is there an issue for this? If not, could you make one? 📓

src/authentication/models/User.ts
oleast: This is a special type, made specifically to handle both users and anonymous users

src/events/api/ruleBundles.ts
oleast: It is that way for all query parameters in Django Rest Framework, unless you specifically opt to change it every time you create a filer class. It wouldn't really make sense for it to be anything else when it is serialized in the url anyways?

src/events/components/CalendarView/CalendarHoverCard.tsx
oleast: Hehe, doing a bit of a shortcut. I'm typecasting away the `undefined`, since this component can't be rendered unless the event is defined

src/events/components/CalendarView/CalendarTile.tsx
oleast: In some places, yes we did. But it was a very special edge case of 'when there is exactly one company connected to the event' thing

src/events/components/EventImage.tsx
oleast: We have our good old boi #45

src/events/components/DetailView/index.tsx
oleast: It is for scrolling to the top of the page when the displayed event is changed. Lets say you go from one event page to another. Keeping the scroll position wouldn't make sense for the user. But react doesn't know that, it just sees a prop change

Kinda just copied it from careeropportunities, since I remember we added it there after an issue was filed on it

src/events/api/ruleBundles.ts
FluidSense: Oh, I just meant the key name, not the type! As I found it weird that event is number, while event could also be an IEvent
`eventId?: number` instead of `event?: number`.

src/events/components/DetailView/index.tsx
FluidSense: Oh, right! So Next doesn't render another "page" and rerenders when the url is changed?

src/events/api/ruleBundles.ts
oleast: This maps directly to the query parameters, so I don't really know how we would do that

FluidSense: Oh, right, now I understand. Perhaps a comment could be added, something alone the lines of "Primary key of event, serialized and named from API" or something?

src/events/components/CalendarView/CalendarHoverCard.tsx
FluidSense: Ah, right! So CalendarHoverCard will only be rendered if the event exists, and thus we can certain that the selector will never return `undefined`?
Cause if not, the next line `const { location, images, title, start_date } = event;` will throw a TypeError.
```js
const { location, images, title, start_date } = undefined;
TypeError: undefined has no properties.
```

src/events/api/ruleBundles.ts
oleast: I think it would potentially be quite tedious to do that for _all_ query parameters which are primary keys?

Could potentially find another way to do query parameters in the future?

src/events/components/CalendarView/CalendarHoverCard.tsx
oleast: Yes, they are used in the calendar, where an event in the calendar is only rendered if it is already in the store

src/events/models/Event.ts
```diff
@@ -79,79 +75,69 @@ export const getEventColor = (n: number): string => {
...
   waitlist: boolean;
   guest_attendance: boolean;
   registration_start: string;
-  unattend_deadline: string;
   registration_end: string;
-  number_of_seats_taken: number;
+  unattend_deadline: string;
   automatically_set_marks: boolean;
-  marks_has_been_set: boolean;
+  rule_bundles: number[];
```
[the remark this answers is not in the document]
oleast: The refactoring of the events API focused a lot on serializing ids instead of inlining related objects

src/events/models/Attendee.ts
```diff
@@ -1,18 +1,22 @@
...
-
-// Attendee from the registration attendee list endpoint.
-export interface IUserAttendee {
+export interface IAttendee {
   id: number;
   event: number;
-  user: IAttendeeUser;
+  user: IUserName;
   attended: boolean;
-  timestamp: IsoDateTime;
```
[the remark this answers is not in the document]
oleast: No, it was just that `IsoDateTime` was an alias for `string` I used in a few places when I created this project and was new to Typescript.

Typescript just infers it to string wherever you use the variable, so it didn't have any effect at all

src/events/hooks/useEventsRepoState.ts
```diff
@@ -15,8 +15,7 @@ const getTextFiltered = (search: string, eventList: IEvent[]) =>
       event.title.toLowerCase().includes(search) ||
       event.description.toLowerCase().includes(search) ||
       event.ingress.toLowerCase().includes(search) ||
-      event.location.toLowerCase().includes(search) ||
-      event.organizer_name.toLowerCase().includes(search)
```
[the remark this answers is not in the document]
oleast: Yes, but that is fixed with #358

src/events/models/RuleBundles.ts
```diff
@@ -1,15 +1,27 @@
...
   rule_strings: string[];
 }
 
-export type GradeRule = number;
+export interface IFieldOfStudyRule {
+  id: number;
+  offset: number;
+  field_of_study: number;
+  field_of_study_display: string;
+}
```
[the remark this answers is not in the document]
oleast: I don't think that should be done in the reducer. Rules can are a used in a lot of different rule-bundles, and doing so in the reducer would just lead to storing multiples of the object.

I think the correct path would be to combine them in selectors

src/events/selectors/event.ts
```diff
@@ -0,0 +1,13 @@
+import { State } from 'core/redux/Store';
+import { eventSelectors } from 'events/slices/events';
+
+export const selectEventCapacity = (eventId: number) => (state: State) => {
+  const event = eventSelectors.selectById(state, eventId);
+  if (!event) {
+    return '-';
+  } else if (!event.is_attendance_event) {
+    return '∞';
```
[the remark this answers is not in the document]
oleast: This is not used in the detail page, and _Dette er ikke et påmeldingsevent_ is still shown there.

This is primarily used for the different front-page views and the event list

src/events/models/RuleBundles.ts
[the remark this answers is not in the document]
oleast: This is also my understanding of the redux style guide: https://redux.js.org/style-guide/style-guide#normalize-complex-nestedrelational-state

src/authentication/models/User.ts
FluidSense: As I've understood them, IAuthUser is based on OW4's implementation of the OIDC-user model, thus extends it.
If this is the standard Django user model, I agree that it fits as it's own type! 🚀

PR #359, opened 2020-04-23
Create basic public attendee list

### LINKED ISSUE #34: Create an an attendee list for attendance events
There should be a list of attendees for events where this is allowed.

No designs have been created yet, and you are free to make what you want!
I would suggest trying it out in the new events [Figma](https://www.figma.com/file/OmjtoRxkGhbarERn2GNvZxBo/Events?node-id=0%3A1), and to make it fit in with the design.

Could probably be implemented as a modal like it is in OW4, or as something more fitting, your choice!

oleast:
Resolves #34 kinda

Creates a sample page with public attendees for each event as a page. This should be integrated as a modal in the event detail view, but since we don't have DS or even a modal yet I made it as a page just to display that it works, and to make it easier to port when we have the things we need.

The design is very basic, and copied from the vent list view.

Screenshot from 2020-04-23 16-39-37


src/pages/events/[eventId]/public-attendees.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+import { PublicAttendees } from 'events/components/DetailView/PublicAttendees';
+
+const PublicAttendeesPage = () => {
+  const { query } = useRouter();
+  const eventId = Number(query.eventId);
+  return <PublicAttendees eventId={eventId} />;
+};
+
+export default PublicAttendeesPage;
```
Comment (FluidSense) [inline src/pages/events/[eventId]/public-attendees.tsx:12]: So this component would perhaps be deleted when integrated with i.e. a modal?
#352 for example?

FluidSense:
How can I test this locally, @oleast?
I tried http://localhost:3000/events/861/public-attendees after logging in, but it is an empty page.

oleast:
> How can I test this locally, @oleast?
> I tried http://localhost:3000/events/861/public-attendees after logging in, but it is an empty page.

My best guess is that it doesn't work because this depends on https://github.com/dotkom/onlineweb4/pull/2471
Not possible to select the attendees for an event if we don't know what event they are connected to :stuck_out_tongue:

src/pages/events/[eventId]/public-attendees.tsx
Comment (oleast) [reply src/pages/events/[eventId]/public-attendees.tsx:12]: Yep

FluidSense:
I see! It works now, yes!
However, I am able to see attendees for events I am not signed up for. Is this intentional?

oleast:
> I see! It works now, yes!
> However, I am able to see attendees for events I am not signed up for. Is this intentional?

Yes, that is intentional. It works that way in OW4 as well

oleast:
We changed it not too long ago 😜

PR #360, opened 2020-04-24
Refactor event calendar so each day gets its own events

### LINKED ISSUE #305: The calendar view of events does not display all events in that month
When selecting the third view for events (calendar) it will not display any of the events that happen on the last day of the month.

In the image linked, there are no events on October 31st, when there should be 3 events.
Screenshot 2019-10-04 at 13 30 42

If you select the second view for events (list) and go back to the calendar view it will display those 3 events as well.
Screenshot 2019-10-04 at 13 30 55

I am not sure if this only happens for the last day of the month.

Link to video https://streamable.com/epgab

### LINKED ISSUE #306: The calendar sorts the events at wrong dates and time
**Describe the bug**
When viewing events in a calendar view it can place events on specific dates, but after half a second re-sort the events based on time. How an event is placed on the wrong date, I am unsure of.

**To Reproduce**
1. Go to 'beta.online.ntnu.no'
2. Click on the calendar view of events (third view)
3. See the events be placed, and perhaps re-sorted

**Expected behavior**
The events should only be placed and sorted once

**Screenshots**
Not a screenshot since it happens too quickly for my mortal hands, but a link to a video: https://streamable.com/p84x0

 - OS: macOS Mojave, 10.14.6
 - Browser: Chrome
 - Version: 77.0.3865.90 (Official Build) (64-bit)

### LINKED ISSUE #337: See events for upcoming month
When at the calendar event view, upcoming month's "offset" is displayed as grey areas.
If there is an event on any of these days, it would be nice to be able to see the event, perhaps as grey as well.
As it is now, there can be confusion as an event can be displayed on e.g. the 03.04 in the April-calendar, but not on the end of March-calendar.

oleast:
Made the calendar a lot simpler by each day just selecting its' own events from the store.
Note that no styling changes have been made as of yet. The styling has just been moved.

Resolved some issues along the way, because they were kinda hard not to fix:
Resolves #337 
Resolves #305 
Resolves #306


src/events/components/CalendarView/CalendarEvent.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface IProps {
+  event: IEvent;
+}
+
+export const CalendarEvent: FC<IProps> = ({ event }) => {
+  return (
+    <Link {...getEventUrl(event.id)}>
+      <a>
+        <HoverCard card={<CalendarHoverCard eventId={event.id} />}>
```
FluidSense, line 20: I feel like you've already explained this to me, but I don't remember the answer.
What is your thoughts about having the event here, passing the id to the CalendarHoverCard then getting the event from the store again from there?

src/events/utils/calendarUtils.ts
```diff
@@ -1,105 +1,95 @@
...
+  /** Set the DateTime to the first day of the month */
+  const firstDay = getFirstDayOfMonth(date);
+  /** Get index of first day from DateTime, zero-index and return */
+  return firstDay.weekday - 1;
+}
+
+/**
+ * @param {DateTime} date Luxon DateTime object.
+ * @returns {number} Length of the given month. Number between 28 and 31.
+ */
```
FluidSense, line 38: Although I'm all in favor of good documentation and comments, is this not a bit superfluous when we have defined the param- and return types through typescript?

oleast: I agree, I didn't really add this now, I just hanged the file so much that it registered changes in all of it :sweat_smile:

src/events/components/CalendarView/CalendarEvent.tsx
oleast: The Redux style guide has a good section on it: https://redux.js.org/style-guide/style-guide#connect-more-components-to-read-data-from-the-store

src/events/utils/calendarUtils.ts
oleast: I ended up removing most of the functions now, since they were not in use

src/events/components/CalendarView/CalendarEvent.tsx
FluidSense: Yes, I agree with the style guide! Their reasoning is that it can prevent rerendering of less DOM-elements by only caring about ID's until the object is actually used to prevent extra rendering.
However, don't we already have the `event` here? Thus, if anything the `event` changes, both `CalendarEvent` and `CalendarHoverCard` will rerender, since both have the `event` itself? If so, fetching it from the store again makes no difference. If the `CalendarTile` instead passed the `eventId` instead of the `event` to this component so that it were similar to the Redux Style Guide example, `CalendarEvent` would not rerender on event change, only the `CalendarHoverCard` would rerender. Seeing as we're using the `event` in both elements, only having the `id` would not work, so the two DOM-elements already are dependent on the same `event`-object and should rerender on the change anyway and can then pass the `event` itself. What do you think?

oleast: I can agree there, and the current solution may not be the best. The initial idea was that the `CalendarHoverCard` was kind of a separate entity.

We can actually solve the re-rendering here in another way, by using minimal values until we actually need something else. We can select just the ids in `CalendarTile`, and select just the title and eventType in `CalendarEvent` to make it work with referential equality. Then pass the eventId to the card

oleast: @FluidSense Had a go at what I mentioned above just now if you want to take a look :camera_flash:

FluidSense: That should definitely deal with the rerendering, indeed!
A very good solution!

Issue #194, opened 2019-02-18
[module] Create pages for company information


oleast:
Should have a page or pages to show off information about companies.

Something similar to the current implementation in [OW4]( https://online.ntnu.no/company/2/), but with a bit more useful information.

Could have things like:
- Past/future events
- Active career opportunities
- Information text about the company
- Contact information and link to website
- How many years they have worked with Online
- Previous names?

Could reuse components from events and career for related features.

oleast:
Should also have more clear links to companies in the event view, pointing the user to information about the company

Should probably also be support for videos to showcase the company

henrikhorluck:
Ref: since how long they have worked with Online, are you thinking we get this information from the oldest event/careeroppurtunity they have, or when they were added to the backend?

oleast:
When they were added would be the easiest to do, and works well enough.

henrikhorluck:
Wouldn't stuff like Gold Sponsor for "fadderuka" 2018, coffeesponsor and so on be stuff we should have on the company-pages as well?

oleast:
Sound like great additions!

I would leave those things out of the initial implementation though.
I think we should get it working with current data first, and expand with more when we can.

mariusaarsnes:
Actually getting some statistics can greatly help Bedkom! 

Everything shouldn't necessarily be visible for everyone but historic data can help the committee with improving events, helping companies improve etc. 

Getting an overview with historic data is something we are looking for in the long run with the new booking system (not part of the MVP).

Issue #50, opened 2018-10-26
Complete Header design


oleast:
The header needs to be completed. Right now it is just a temporary design, created to have *something*.
Currently, it does not scale at all for mobile devices.

@sardred Has looked into designing this earlier.

Tasks:
- [ ] Color?
- [x] How to highlight the page you are currently on.
- [x] How to do the login functionality.
- [x] Make it scale.
- [x] Decide on how to display the many sub-pages. There is quite a lot to show.

oleast:
I think that since we are going with a sort of app-like experience for mobile users, it would be nice to go with a bottom style menu for our mobile version?

Having a few important icons available in the menu at all times (like 4 or 5), and leaving the rest of the options in a side-drawer type thing?

mariush2:
Do we still feel the need for a bottom style menu? I can't really decide if its necessary, do we have any examples of other sites using this sort of bottom menu?

niklasmh:
I think a bottom menu is a good idea as it makes the menu more accessible. Android apps, like twitter and facebook, have also recently (1-2 years ago) made such a transition. Apple apps mostly had it from the beginning - and it seems like it was a good idea as it did not go the other way.

Personally, I would prefer that web apps did the same (as of tablet-sized phones), but by doing this you create a totally different layout for desktop and mobile. This may lead to inconsistencies and more effort on the development side.

You can look at how https://m.youtube.com has implemented it. They have kept it small with four buttons. They also have a top menu. They also hide both the top and bottom menu when scrolling down - which may solve the problem with using too much space on the screen.

This is a nice to have, and I do not see any disadvantages weighing out the pros. But reasoning from the current situation of dotKom, we should probably focus on finishing the design overall first. Meanwhile, we can add this to the backlog.

PR #365, opened 2020-04-26
Add company pages

oleast:
For now just created some similar styling to the `hobbygroups` pages to make company cards for a list

Depends on https://github.com/dotkom/onlineweb4/pull/2473 for data

Resvoles #194


oleast:
Some things are still missing, but I would still appreciate a review already :smile: 

Still missing:
- [] Finishing touches on styling for company detail page
- [] Some actual search functionality for the company list page

There should also be some way to link to the company list page, but that is part of a larger problem of linking to all our services #50

src/companies/components/CompanyDetail/CompanyCareerOpportunities/CareerOpportunity.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const CareerOppotunity: FC<IProps> = ({ oppotunity }) => {
+  const deadline = DateTime.fromISO(oppotunity.deadline).toLocaleString();
+  return (
+    <Link {...getCareerOpportinityUrl(oppotunity.id)}>
+      <a className={style.container}>
+        <p className={style.name}>{oppotunity.employment.name}</p>
+        <p className={style.count}>{oppotunity.title}</p>
+        <p>{oppotunity.location.map((location) => location.name)}</p>
+        <p className={style.count}>{deadline}</p>
```
henrikhorluck, line 22:
```suggestion
        <p className={style.count}>{deadline || "Ikke spesifisert"}</p>
```

src/companies/components/CompanyList/NextPageObserver.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return state.companies.search.loading === 'pending';
+};
+
+const selectTotalCount = () => (state: State) => {
+  return state.companies.search.count;
+};
+
+const selectCurrentCount = () => (state: State) => {
+  return state.companies.search.ids.length;
+};
```
FluidSense, line 47: This observer, is this what you talked about which would allow for smoother population and rerending?

oleast: It's been a while since I wrote this, I can't recall what you are referring to?

oleast:
> 1. There's an awful lot of API-calls for searches. Could we implement some kind of cache so that only "new" queries for a session is called? So that "s", "se", "s" would be reduced to 2 calls instead of 1, since "s" has already been queried and should be in the story? Perhaps with some timestamp to allow for requerying if you have been on the page for a long time?

That could be possible to solve in some way. Caching is a huge thing in and of itself, and we should look into default caching for `fetch` requests. It can be done in multiple ways, but I think there is built in support for it in ServiceWorkers?

> 2. Could subheaders be more prominent, perhaps bolded?
>    I found it a bit difficult to notice the sections "Arrangementer" and "Aktive jobbannonser".

I agree on that one. The page here is just rendered markdown just as all other pages that use it. The font becoming this small is because the inclusion of Design System. Could potentially be solved by using the markdown renderer from that project?

src/companies/components/CompanyList/NextPageObserver.tsx
FluidSense: Let me rephrase: What exactly is a NextPageObserver? Is it a standard Next-thing?

oleast: Ah, no. It's a thing I wrote to listen for when the next "_page_" of an infinite scrolling list should be requested

PR #366, opened 2020-05-02
Handle missing Offline images

FluidSense:
Although Offline-images should be present, OWF should be resistant to faults 💪

### DIFF
--- a/src/frontpage/models/Offline.ts
+++ b/src/frontpage/models/Offline.ts
@@ -6,5 +6,5 @@ export interface IOfflineIssue {
   issue: string;
   release_date: string;
   title: string;
-  image: IResponsiveImage;
+  image?: IResponsiveImage;
 }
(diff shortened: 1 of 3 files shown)


src/frontpage/models/Offline.ts
```diff
@@ -6,5 +6,5 @@ export interface IOfflineIssue {
   issue: string;
   release_date: string;
   title: string;
-  image: IResponsiveImage;
+  image?: IResponsiveImage;
```
oleast, line 9: I think this should be `IResponsiveImage | null` instead of possibly `undefined`. The API will never return undefined, as JSON does not support `undefined`, and all `None` values in Python are serialized to `null`. 

Very minor thing, but it might come back and bite us in the ass if we ever do a strict equality check instated of a a truthy/flasy check like above

PR #370, opened 2020-05-03
Fix oidc app failing to fetch clients in app connections

Aborysa:
- Add '/' to end of client request to prevent redirect
 - Remove Authorization header if no user is supplied to performRequest

### DIFF
--- a/src/common/utils/api.ts
+++ b/src/common/utils/api.ts
@@ -37,8 +37,12 @@ const performRequest = async (query: string, parameters: IQueryObject = {}, opti
   }
   const headers = {
     ...options.headers,
-    Authorization: options.user ? `Bearer ${options.user.access_token}` : '',
-  };
+  } as Record<string, string>;
+
+  if (options.user) {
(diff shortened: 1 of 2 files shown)


Aborysa:
https://img.spaceway.network/chrome_030520_1609_7039d9737a.png

src/common/utils/api.ts
```diff
@@ -39,6 +39,12 @@ const performRequest = async (query: string, parameters: IQueryObject = {}, opti
     ...options.headers,
     Authorization: options.user ? `Bearer ${options.user.access_token}` : '',
   };
+
+  // requests sometimes fail to resolve without error when Authorization is set to ''
+  if (!options.user) {
+    delete headers.Authorization;
+  }
+
```
oleast, line 47: These statements sort of contradict each other, as we are always adding something (the Authorization header), and then removing it in some cases afterwards.

Maybe we should just do `if options.user`; add `Authorization` header?

Aborysa: sure

PR #371, opened 2020-05-07
Create a new Header Design (featuring authentication in redux and silent signin)

### LINKED ISSUE #50: Complete Header design

### LINKED ISSUE #317: Automatically check onlineweb4 user login state
Right now you need to explicitly click on login every time the access token expires. 

The app should check if the user is logged in to onlineweb4 and if the user is, automatically fetch an access token. This makes it so the only time the login button is ever present is when you're actually not logged in to the system at all.

oleast:
This is very much a work in progress, and I don't actually know how I want it to look in the end. I'm trying things out, and hope we can go for something.

After creating this PR I found out that it kinda works a bit like the header of 'prisguiden.no', which I don't actually have a problem with. Maybe we could even make it more like that?

This PR covers a lot of things, and maybe not all of i should be mixed, but it happened, and we're kinda stuck with it. There are changes in a lot of files, but most of it is just using a different way of getting the current user (using OIDC methods instead of getting the user from react state).

Resolves #50 
Resolves #317


src/core/components/PageHeader/PageHeader.less
```diff
@@ -0,0 +1,95 @@
...
+  max-width: 1200px;
+  margin: auto;
+  justify-content: space-between;
+}
+
+.mainNavItems {
+  display: flex;
+  justify-content: space-evenly;
+  align-items: center;
+  height: 100%;
```
henrikhorluck, line 43:
```suggestion
```
This makes the text vertically-centered

oleast:
Thoughts for the future:
- Create a different front page for authenticated and non-authenticated users.
- Move statistics out of profile
- Move user search out of profile
- Move marks out of settings

PR #381, opened 2020-09-10
Attend events

FluidSense:
Note: This does not check whether you are logged in before displaying the attend button. 
Also, you cannot unattend.
Also, you cannot pick extras.
🤫

### DIFF
--- a/src/events/components/AttendButton/AttendButton.tsx
+++ b/src/events/components/AttendButton/AttendButton.tsx
@@ -1,23 +1,18 @@
-import { FC, useState, useEffect } from "react";
-import Button from "core/components/errors/NotAuthenticated/Button";
-import CaptchaModal from "./CaptchaModal";
-import { DateTime } from "luxon";
-import { getAttendeeForEvent, userAttendEvent } from "events/api/attendee";
+import { FC, useState, useEffect } from 'react';
+import Button from 'core/components/errors/NotAuthenticated/Button';
+import CaptchaModal from './CaptchaModal';
+import { DateTime } from 'luxon';
+import { getAttendeeForEvent, userAttendEvent } from 'events/api/attendee';
 
 interface IAttendButtonProps {
-  eventId: number
+  eventId: number;
   registrationStart: DateTime;
   registrationEnd: DateTime;
   unattendDeadline: DateTime;
 }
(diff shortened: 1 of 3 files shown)


src/events/components/AttendButton/AttendButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [attending, setAttending] = useState<boolean>(false); // Fetch is attending
+  const currentTime = DateTime.local(); // Check if this works
+
+  useEffect(() => {
+    const fetchAttending = async () => {
+      const attendingFetch = await getAttendeeForEvent(eventId);
+      setAttending(attendingFetch.attended);
+    };
+    fetchAttending();
+  }, [eventId]);
```
kharann, line 26: Not sure if we need eventId on the useEffect. It should fetch only once anyway since you need to change url to go to another event?

src/events/components/AttendButton/AttendButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+      <Button onClick={signOff}>Meld meg av</Button>
+    );
+  if (currentTime > registrationStart)
+    return (
+      <div>
+        <Button onClick={toggleModal}>Meld meg på</Button>
+        {modal}
+      </div>
+    );
+  if (currentTime > unattendDeadline || !attending) return <p>cant unattend</p>; // cant unattend
```
kharann, line 52: Engelsk, kanskje dette skal være norsk?

kharann (approved):
Looks pretty good so far. Maybe we should create an utility file which helps us with the different case logics. I do believe we will have many more cases to come which affects if the button should be shown or not.

Like we have cases for when the user is logged in, logged out, matching the rule_bundles, deadlines, extras etc.

We could probably make it another issue... probably kek

src/events/components/AttendButton/AttendButton.tsx
oleast: Changing the URL is what triggers the eventId to change though.
If you were to go from one event page to another, I'm you would still have the data from the first event page you visited.
This is why all dependencies should always be in the array.

FluidSense:
resolved #352

PR #384, opened 2020-09-11
Refresh page after attending

kharann:
We have some iffy bugs where the attending button should be "Meld av" after attending, and the number of attendees is not counting with yourself!


FluidSense (approved):
Hahaha, ouff, but lets go!

PR #386, opened 2020-09-11
Conditional attend button

FluidSense:
This will show a text to users who cannot attend an event for some reason.
It will also display a defective "Meld meg av"-button


kharann (approved):
Looks pretty good to me. I do think we maybe should make this a switch case with enums eventually because we have so many cases.

src/events/components/DetailView/AttendanceEvent.tsx
```diff
@@ -23,10 +23,6 @@ const AttendanceEvent: FC<IProps> = ({ eventId }) => {
   // @ts-ignore
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
oleast, line 24: Are these still needed?

FluidSense: Nope! Removing them now!

FluidSense:
> Looks pretty good to me. I do think we maybe should make this a switch case with enums eventually because we have so many cases.
Most cases are handled in OW4 with a status message! Looks way cleaner now imo

PR #388, opened 2020-09-11
Feature/old ow toast

Paalar:
oh alsoupgrade to 1.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "typescript": "3.8.3"
   },
   "dependencies": {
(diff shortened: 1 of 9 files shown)


Paalar:
Looks good to me

FluidSense (approved):
I did this as well

oleast:
Would this not be more of an "Alert" thing than a toast?

FluidSense:
> Would this not be more of an "Alert" thing than a toast?

You're probably right!

PR #405, opened 2020-09-12
Add a Link which takes you to the wallet page

kharann:
image

### DIFF
--- a/src/core/components/Header/Login.tsx
+++ b/src/core/components/Header/Login.tsx
@@ -3,7 +3,7 @@ import React, { Component, ContextType } from 'react';
 import LoginView from 'authentication/components/Login';
 import { IAuthUser } from 'authentication/models/User';
 import { UserContext } from 'authentication/providers/UserProvider';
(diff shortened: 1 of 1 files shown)


FluidSense:
What are your thought on having Saldo in the dropdown menu vs. at the profile page?

kharann:
> 
> 
> What are your thought on having Saldo in the dropdown menu vs. at the profile page?


I didn't feel there was a spot in profile which made sense to have wallet. It's also its own page, so it should just be a redirect link, rather than pages directly under profile. You could maybe move it to profile, or not

FluidSense:
In my mind, you add money to your account, and thus is a part of the profile/account. This might just be how I think about it because that was how OW4 did it. The page's content could be probably put as a subpage of the profile as well. But I agree that for now, since it is it's own page, it can be a link right there.

FluidSense (approved):
yup yup 👍

kharann:
I agree moving it to a subpage in the future

PR #417, opened 2020-09-14
Fix small article css

Paalar:
I believe there's either some crazy div inheriting I can't find or some weird bug, which makes the `<div className={style.smallContainer}>` get the style `style.articleContainer` as the div gets the hover effect from it. But get this, it loses it on the second render (i.e. if you go to an event, and go back, it renders it correctly).
An easy fix was to wrap it inside of a `span` instead.


FluidSense (approved):
LGTM, weird fix to a weird problem

PR #420, opened 2020-09-14
change to network then cache

aslakhol:
This PR makes some changes to our serviceworker and the caching strategies outlined in it. 

If this is merged we will be using a network-then-cache strategy. Which means that when a fetch is made we send that request of to the server. The response is copied and used both to update the cache and served to the user. Cache is only used as a fallback in case of no network. 

This is slower then a cache first strategy, where you reply with the response from cache right away. However this also ensures that the user get the most recent data, while ensuring partial functionality while offline. 

We could employ a much more complex caching strategy. Precaching certain routes on install or using different approaches for the different kinds of information we provide, but this is a reasonable first step. 

These two links are excellent resources for understanding the potential: 
https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/
https://developers.google.com/web/fundamentals/primers/service-workers

However there is also a definite question to be asked surrounding return on investment. At the moment I don't think any of the potential changes we can make are critical, and as such I suggest letting this rest unless someone is very motivated to learn about offline-first.

### DIFF
--- a/public/serviceworker.js
+++ b/public/serviceworker.js
@@ -1,12 +1,43 @@
 const sw = self;
+const cacheName = 'owf-cache';
+const cachesToKeep = [cacheName];
 
-sw.addEventListener('install', (event) => {
-  event.waitUntil(caches.open('base-cache').then((cache) => cache.add('/')));
+sw.addEventListener('install', () => {
+  self.skipWaiting(); // Ensures that any new, waiting eventlistener becomes the active one.
 });
 
+// Cleanup unwanted caches
+self.addEventListener('activate', (event) => {
+  event.waitUntil(
+    caches.keys().then((cacheNames) => {
+      return Promise.all(
+        cacheNames.map((cache) => {
+          if (!cachesToKeep.includes(cache)) {
+            return caches.delete(cache);
+          }
+        })
+      );
+    })
+  );
+});
+
+// Respond to fetches with a network then cache strategy.
 sw.addEventListener('fetch', (event) => {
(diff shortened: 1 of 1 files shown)


public/serviceworker.js
```diff
@@ -1,13 +1,44 @@
...
+const cachesToKeep = [cacheName];
 
-sw.addEventListener('install', (event) => {
-  event.waitUntil(caches.open('base-cache').then((cache) => cache.add('/')));
+sw.addEventListener('install', () => {
+  self.skipWaiting(); // Ensures that any new, waiting eventlistener becomes the active one.
 });
 
+// Cleanup unwanted caches
+self.addEventListener('activate', (event) => {
```
FluidSense, line 10:
```suggestion
sw.addEventListener('activate', (event) => {
```

public/serviceworker.js
```diff
@@ -1,13 +1,44 @@
 const sw = self;
+const cacheName = 'owf-cache';
+const cachesToKeep = [cacheName];
 
-sw.addEventListener('install', (event) => {
-  event.waitUntil(caches.open('base-cache').then((cache) => cache.add('/')));
+sw.addEventListener('install', () => {
+  self.skipWaiting(); // Ensures that any new, waiting eventlistener becomes the active one.
```
FluidSense, line 6:
```suggestion
  sw.skipWaiting(); // Ensures that any new, waiting eventlistener becomes the active one.
```

FluidSense (approved):
Coool 🧊 🐧 ⛸️

PR #421, opened 2020-09-14
Calendar today CSS

Paalar:
Make today span inline block, to better fit background color and have same height as other views

## Old

[
Screenshot 2020-09-14 at 21 15 20
](url)

## New

Screenshot 2020-09-14 at 21 15 29


kharann (approved):
🎱  says this looks pretty good to him

Issue #409, opened 2020-09-13
Forgotten password


FluidSense:
There is currently a button for "Forgotten password" which does nothing. 
This feature could be implemented as a page with an input field and a SEND-button that sends a `POST` to `online.ntnu.no/auth/recover` with a` x-www-form-encoded` form with `"email"`: user-entered e-mail.

I'm not certain whether or not a CSRF-token has to be included. If so, this might be a bit more work and might require changes to the OW4-API.

Another issue is that the response is a 302 Redirect with the message of success/not success placed in a header named "Set-cookie", with JSON data.

FluidSense:
It now does something, but the magic links redirects to "/" and thus does not work

PR #424, opened 2020-09-15
Reset password & register as links to OW4

FluidSense:
Currently, the "Register" and "Reset password"-buttons does nothing. 
I figured they could point to the functionality for those things in OW4.
I kinda think that is the correct way to do it, since the OIDC-authority is OW4. 
One could however argue that since OWF is OW4's frontend, it should be included here as well and replace the old methods for this functionality at OW4, such as I do in #409

Note: As relative URLs are used, one have to add `https://online.ntnu.no/` to the a-tags to test locally.
Also note: If you _actually_ are signed in to OW4 and tries to use the buttons, it will silently fail as the error message is inserted as a header in the 302 Redirect. 
I.e. if you press 'Register' while you have an active session at OW4, but no active token at OWF, you will be redirected to '/' with no error message displayed.


kharann (approved):
LGTM.

PR #437, opened 2020-09-18
fixes toast hell :fire:

s0lvang:


aslakhol (approved):
I'm ashamed to have review this.

PR #438, opened 2020-09-18
End session on OW4 on logout 

FluidSense:


FluidSense:
This works to actually log you out, but the `post_redirect_url` parameter is not honored and users are redirected to the login page. (With a message saying they are successfully logged in??).
I have a hunch that this is due to the package at OW4, but I am currently troubleshooting it.

FluidSense:
I am pretty sure this is the correct way, and that the weird redirect exists at OW4 and as such this could be merged

s0lvang (approved):

Issue #427, opened 2020-09-16
Reverse the list of events in the middle tab on the landingpage


Westermoen:
The list of events shown in the 2nd/middle tab on the landingpage displays the events in the wrong order as shown in the image below.
The list should be reversed so the event occuring closest to the present day is shown on top.

<img width="1132" alt="Skjermbilde 2020-09-16 kl  10 18 04" src="https://user-images.githubusercontent.com/15379479/93311156-ec3a7380-f805-11ea-8dc6-77434cc9350f.png">

FluidSense:
Reversing this view without any further changes will also reverse the view in /events, resulting in wrong ordering on infinite scroll-loading. The sort order could probably be decided by a prop so the views can sort differently.

FluidSense:
The events are also sorted in the Redux-store, and as such require a manual sort after retrieving the events from the store.

PR #441, opened 2020-09-20
fix sortOrder in listView

kaffikaffi:
Added a prop in EventListComponent, such that you can decide the the order you want to sort the events in. The default value is descending, so the frontpage will have the correct sorting by default.

image


kaffikaffi:
resolves issue #427 👍

kharann:
:8ball: says this is good stuff

kharann (approved):

PR #442, opened 2020-09-20
add apply button on career page

kaffikaffi:
There was no apply button on the front-end, but it was one on old OW. So I added one.


kharann (approved):
:8ball:  says this is some :fire:  stuff

PR #445, opened 2020-09-20
Small redesign of carreer list page

kharann:
I added some rounded edges to the career page
image

<details>
<summary>Before</summary>
before

</details>
<details>
<summary>After</summary>
after
</details>


FluidSense (changes requested):
I think this PR could be better with some sharp edges

kharann:
> I think this PR could be better with some sharp edges

This looks super M O D E R N. I swear

PR #475, opened 2020-09-27
Server side authentication

FluidSense:
Change from `oidc-client` to `next-auth` to allow for server side rendering. 
This is motivated by a change[ coming to browsers which disallow iframe renewal of tokens](https://leastprivilege.com/2020/03/31/spas-are-dead/), and as such users will be logged out often.
By using Next as a backend-for-frontend, Next can keep renew tokens and keep secrets, allowing for safe renewal of tokens through other flows than PKCE.

## To-do before undraft:

- [ ] Fix types on `pages/api/auth`
- [ ] Fix types on `useSession()`'s returned user so that it can share a type with the one actually in use at `pages/api/auth` and not need `(session.user as unknown as IAuthUser).profile.` this to cast it every time it is needed.


balazsorban44:
NextAuth.js maintainer her! Gi meg en pip hvis dere trenger hjelp med dette. :wink:

Mye som har endret seg siden `3.1.0`, vi har en v4 beta nå også (https://next-auth.js.org/getting-started/upgrade-v4)!

Innebygd TypeScript, bedre docs, enklere/mer logiske signaturer, osv.

PR #480, opened 2020-09-29
Show attendee button

Paalar:
Simple button to show the attendees if the user is attending as well. 
The attendees are shown in a modal.

Yes i am aware of the branch spelling.

Screenshot 2020-10-01 at 14 38 06
Screenshot 2020-10-01 at 14 38 13
Screenshot 2020-10-01 at 14 38 23
Screenshot 2020-10-01 at 14 40 17


src/events/components/DetailView/PublicAttendees/index.tsx
```diff
@@ -1,29 +1,46 @@
...
 }
 
-export const PublicAttendees: FC<IProps> = ({ eventId }) => {
+const attendeeList = (attendees: IPublicAttendee[]) => (
+  <div className={style.grid}>
+    {attendees.map((attendee, index) => (
+      <Attendee key={attendee.id} attendee={attendee} count={index + 1} />
+    ))}
+  </div>
+);
```
oleast, line 23: Why is this a function and not a component?

src/events/components/DetailView/PublicAttendees/index.tsx
```diff
@@ -1,29 +1,46 @@
...
   const attendees = useSelector(selectPublicAttendeesForEventId(eventId));
+  const [showModal, setShowModal] = useState(false);
+
+  const toggleModal = () => setShowModal(!showModal);
 
   useEffect(() => {
     dispatch(fetchPublicAttendeesByEventId(eventId));
   }, [eventId]);
 
+  if (!isAttending) return null;
```
oleast, line 36: Why was this moved to inside the component?
Would that not result in the fetch function being called even when it shouldn't?

src/events/components/DetailView/PublicAttendees/index.tsx
```diff
@@ -1,29 +1,46 @@
...
 import { fetchPublicAttendeesByEventId } from 'events/slices/publicAttendees';
 
 import { Attendee } from './Attendee';
 import style from './PublicAttendees.less';
+import { IPublicAttendee } from 'events/models/Attendee';
+import { Button, Modal } from '@dotkomonline/design-system';
 
 interface IProps {
   eventId: number;
+  isAttending: boolean;
```
oleast, line 14: If I recall correctly the rules for showing attendees should be that you are either eligible for signup or attending the event.

Looks like the API doesn't actually restrict it, which it should. https://github.com/dotkom/onlineweb4/blob/develop/apps/events/api/views.py#L125

Current view logic in OW4 can be found her: https://github.com/dotkom/onlineweb4/blob/develop/templates/events/details.html#L400
(where `user_status.status` is a boolean that refers to if the user is eligible for signup I think?)

src/events/components/DetailView/PublicAttendees/index.tsx
Paalar: I started it of as a function component, but then switched back to a function. It is just personal preference, but I did it because:

- When only passing 1 variable, I don't think it matters much (visually) to write `{attendeeList(attendees)}` compared to <AttendeeList attendees={attendees} />
- When using only 1 variable, I feel it is redundant to define the props. Writing an interface creates 3 more lines of text and a couple of lines of white space
- I could also define the props inline, but that makes it harder to read what is happening at first glance, which I am not that much of a fan of.

I can change it back, if it is wanted, but it was just for my personal (visual noise) reasons.

src/events/components/DetailView/PublicAttendees/index.tsx
Paalar: This is fair. I moved it inside since it would make the parent file less cluttery. Either then creating another component that does the null check or writing a conditional render, or another soluton.

src/events/components/DetailView/PublicAttendees/index.tsx
Paalar: That should be a quite easy fix!

Paalar:
Currently struggling to create a button that will put your status as 'visible' on the waiting list.

When checking out `https://online.ntnu.no/api/v1/docs/` I found `/api/v1/event/attendees/{id}/change/` which seemed perfect, but I have some troubles with it.
When I try to perform the request through Postman, I get the error `"detail": "Metoden \"GET\" ikke gyldig."`, which is super weird, since I'm using PUT/PATCH as described on Swagger.
I eventually got so annoyed with the response, so I tried to implement the request in OWF. The response I get there is just a plain old 404.

Any suggestions or tips would be appreciated.

src/events/components/DetailView/PublicAttendees/index.tsx
Paalar: This has now been moved into its own component and logic.

src/events/components/DetailView/PublicAttendees/index.tsx
Paalar: Has been fixed by adding the `canAttend` prop and check if null (not logged in) and it's status (true if the user eligible).

src/events/components/DetailView/PublicAttendees/index.tsx
```diff
@@ -1,46 +1,14 @@
...
-  return (
-    <>
-      <Button onClick={toggleModal}>Vis påmeldte</Button>
-      <Modal open={showModal} onClose={toggleModal}>
-        {attendeeList(attendees)}
-      </Modal>
-    </>
-  );
+export const PublicAttendees: FC<IProps> = ({ eventId, isAttending, canAttend }) => {
+  if (!canAttend) return null; // The user is not logged in
```
Paalar, line 11: I guess I can do the 3 boolean checks in one if, but I felt it was reasonable to not do them all in one (harder to read) and it makes sense in an interaction aspect to check if the user is logged in first.

aslakhol (approved):

PR #485, opened 2020-09-30
Change CI to github actions

kharann:


kharann:
Probably just squash it. I might use this later with deployment to azure

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
+name: Continuous Integration
+on:
+  push:
+
+jobs:
+  run-linters:
+    name: Run linters
```
aslakhol, line 7: Everything is in one job called run-linters.

kharann: The action runs them in sepperate jobs. So in total 3 different jobs will be instanced during this

oleast:
Are the steps for `compile typescript` and `prettier` missing?

kharann:
> Are the steps for `compile typescript` and `prettier` missing?

I do believe eslint will run prettier because of the prettier plugin in the eslint config (I also tested this and it will throw errors on prettier errors).

Compile typescript just runs eslint I believe

oleast:
> I do believe eslint will run prettier because of the prettier plugin in the eslint config (I also tested this and it will throw errors on prettier errors).

I'm unsure about this, but it may be.

> Compile typescript just runs eslint I believe

In my experience they will pick up different errors.

kharann:
> > I do believe eslint will run prettier because of the prettier plugin in the eslint config (I also tested this and it will throw errors on prettier errors).
> 
> I'm unsure about this, but it may be.
> 
> > Compile typescript just runs eslint I believe
> 
> In my experience they will pick up different errors.

I took a look at drone and saw it only runs npm run compile-tsc which basically just runs eslint. Am I misunderstanding?

I'll add prettier back to the linter tho

oleast:
> I took a look at drone and saw it only runs npm run compile-tsc which basically just runs eslint. Am I misunderstanding?

Not really. `npm run tsc` will run `tsc --noEmit`, which does a typecheck with the typescript compiler. It will error whenever types are wrong, mismatched or missing.

Eslint will check the code style and usage, but not that the actual types are sound.

There is a small piece of overlap, such as both triggering on unused variables, but they each check their own things as well.

oleast (approved):

Issue #483, opened 2020-09-30
Replace temporary Online-icons with official logo.


PeterRydberg:
**Describe the bug**  
Not much of a _bug_. The (very cool!) icons on OWF are wrong, and should be replaced by Online's official logo now that the new frontend is launched in production. This is to prevent wrong logo associations and confusion. Should fix itself by replacing the files in `public/img/icons/` ⚡.

**To Reproduce**
1. Look at favicon in the tab.
2. Look at the background on empty event images.
3. Possibly more places? May need some extra insight here.
4. These icons are not from the official Online logo.

**Expected behavior**
The redesigned logo should be used as standard where needed.

**Screenshots**
![image](https://user-images.githubusercontent.com/22096301/94687177-2d657400-032c-11eb-8fb6-4a78914ac0b0.png)
![image](https://user-images.githubusercontent.com/22096301/94687187-32c2be80-032c-11eb-8b90-8153a45d070f.png)
etc.

oleast:
R.I.P my logo 💀

FluidSense:
@PeterRydberg I have read the "grafisk profil"-page, but I am still uncertain: Can we use the logo-icon with the current colors? The current colors makes it easy to discern the type of event when an image is not present

FluidSense:
Also, @dotkomonline/design-system ships both the current logo and icon, and can be used instead

PeterRydberg:
You may! I might clarify that page a bit, but seeing as it has a practical purpose, changing the colors for _icon_ use rather than _logo_ use it should be permitted.

Paalar:
I believe #487 solves this issue.

PR #487, opened 2020-10-01
Change online icons to current logo

Paalar:
As #483 states, the current icons are wrong.

- [x] New PWA icons that match the current logo
- [x] Change of the favicon
- [x] Change other icons that imitate the logo to the current logo. [Awaiting this PR that makes color props optional](https://github.com/dotkom/design-system/pull/188)

I believe I have changed all the uses of the wrong logo to the correct one. The articles and magazines do not use the logo as it should. You can see it in the images below where I turned off the wi-fi. 
Screenshot 2020-10-02 at 15 53 23

Screenshot 2020-10-02 at 15 53 16

A preview of the new PWA icons. They use the proper logo from online.ntnu.no

pwa-icon-v1-192


public/manifest.webmanifest
```diff
@@ -9,12 +9,12 @@
       "type": "image/png"
```
oleast, line 9: Looks like the icon above () which I can't tag at is not replaced?

oleast (approved):
Looks good to me 😄

oleast (changes requested):
Ok maybe not

owf icons

Paalar:
> Ok maybe not
> 
> owf icons

Haha whoops 🤡

public/manifest.webmanifest
Paalar: The PWA icon? I thought it would be changed when the image itself was replaced?

Paalar: Ah you mean `icon-256.png`. I got a bit confused. It has been updated, but since it's the same name, it doesn't show as deleted, just changed. You can see that it is not the same by seeing ` BIN -23.5 KB (32%)`. 

Also included a screenshot of it. It is the image that is used as a favicon, so you can see that it is not the same in the Vercel preview.
Screenshot 2020-10-02 at 17 05 09

oleast: Sorry about that one, I thought the image had been deleted 😅 
deleted images

The only difference I can see here is the percentage visible on the replaced icon. Didn't look close enough 😛

oleast (approved):
Nice work!

If you'd like you can set the icon color to be the slightly more faded version, but I'll leave that up to you ☺️ 
faded

Paalar:
> Nice work!
> 
> If you'd like you can set the icon color to be the slightly more faded version, but I'll leave that up to you ☺️

I agree that it makes the icons fit better on the screen!

PR #493, opened 2020-10-05
Update Redux-store events if newer is retrieved

### LINKED ISSUE #481: Number of attendees not updated if you entered the page with direct link to event
**Describe the bug**  
Number of attendees is not updated on the frontpage if you entered with a direct link to the event

**To Reproduce**
1. Go to [https://online.ntnu.no/events/993](https://online.ntnu.no/events/993)
2. Click on Online logo to go to home screen
4. See error that number of attendees is not updated for the event you was visiting, it shows 0/XX even if it should be 10/XX
5. Refersh with Ctrl+Shift+R and the correct number is loaded

**Expected behavior**
See correct number of attendees

 - OS: Windows
 - Browser: Chrome & Firefox

**Additional context**
This is probaly related to caching and this issue: [434](https://github.com/dotkom/onlineweb-frontend/issues/434)

FluidSense:
Fixes #481 
`addMany` does not do a deep similarity check, and thus does not update the list of events, which then looks confusing when an event is viewed directly, but then not updated in the main view. This is then done with `upsertMany` (update if present else insert = upsert)

### DIFF
--- a/src/events/slices/events.ts
+++ b/src/events/slices/events.ts
@@ -178,7 +178,7 @@ export const eventsSlice = createSlice({
     builder.addCase(fetchEvents.fulfilled, (state, action) => {
       state.loading = 'idle';
       const events = action.payload.results;
-      eventsAdapter.addMany(state, events);
+      eventsAdapter.upsertMany(state, events);
     });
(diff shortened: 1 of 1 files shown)


kharann (approved):
LGTM

PR #497, opened 2020-10-06
Confirm your attendance/unattendance with a toast message

kharann:
small cute popup when unattending/attending
unattend

### DIFF
--- a/src/events/components/Attendance/AttendButton.tsx
+++ b/src/events/components/Attendance/AttendButton.tsx
@@ -23,7 +23,7 @@ const AttendButton: FC<IAttendButtonProps> = (props: IAttendButtonProps) => {
         await dispatch(setAttendeeByEventId({ eventId, captcha: token })).then(unwrapResult);
         addToast('Du har blitt meldt på arrangementet');
       } catch (err) {
-        addToast('Noe gikk galt under påmeldelse av arrangement', { type: 'error' });
+        addToast(`Noe gikk galt under påmeldelse av arrangement, ERROR: ${err.message}`, { type: 'error' });
       }
     }
   };
--- a/src/events/components/Attendance/AttendButton.tsx
+++ b/src/events/components/Attendance/AttendButton.tsx
@@ -5,6 +5,7 @@ import { setAttendeeByEventId } from 'events/slices/attendees';
 import { Button } from '@dotkomonline/design-system';
 import { useToast } from 'core/utils/toast/useToast';
 import { unwrapResult } from '@reduxjs/toolkit';
+import style from './attendance.less';
 
 interface IAttendButtonProps {
   eventId: number;
@@ -31,7 +32,9 @@ const AttendButton: FC<IAttendButtonProps> = (props: IAttendButtonProps) => {
 
   return (
     <>
-      <Button onClick={toggleModal}>Meld meg på</Button>
+      <Button color="success" onClick={toggleModal} className={style.button}>
+        Meld meg på
+      </Button>
       {modal}
     </>
   );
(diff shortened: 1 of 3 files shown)


FluidSense:
Neato! What about changing the color to `color="secondary"` on the "Meld meg av"-button as well? To indicate that this action is not the same as the previous action?

kharann:
> Neato! What about changing the color to `color="secondary"` on the "Meld meg av"-button as well? To indicate that this action is not the same as the previous action?

Perhaps...

src/events/components/Attendance/UnattendButton.tsx
```diff
@@ -11,7 +12,11 @@ interface IAttendButtonProps {
 
 const UnattendButton: FC<IAttendButtonProps> = ({ eventId, isOnWaitList, waitListNumber }) => {
   const dispatch = useDispatch();
-  const signOff = () => dispatch(removeAttendeeByEventId(eventId));
+  const [addMessage] = useToast({ type: 'success', duration: 5000 });
+  const signOff = async () => {
+    await dispatch(removeAttendeeByEventId(eventId));
+    addMessage('Du har blitt meldt av arrangementet');
```
FluidSense, line 18: Can we somehow check if the attendee _actually_ was removed? In case things fail?

kharann:
Will display error if there's an error field from dispatch

failed_unattend

src/events/components/Attendance/AttendButton.tsx
```diff
@@ -12,18 +15,26 @@ const AttendButton: FC<IAttendButtonProps> = (props: IAttendButtonProps) => {
...
 
-  const signUp = (token: string | null) => {
+  const signUp = async (token: string | null) => {
     if (token) {
-      dispatch(setAttendeeByEventId({ eventId, captcha: token }));
+      try {
+        await dispatch(setAttendeeByEventId({ eventId, captcha: token })).then(unwrapResult);
+        addToast('Du har blitt meldt på arrangementet');
+      } catch (err) {
+        addToast(`Noe gikk galt under påmeldelse av arrangement, ERROR: ${err.message}`, { type: 'error' });
```
FluidSense, line 26: Since you are already using `.then()`, could perhaps the `try-catch` be minified to `.then().catch()`? And if you are using `.then()`, could we remove the `await` keyword here by chaining everything with `.then()`?

src/events/components/Attendance/UnattendButton.tsx
```diff
@@ -11,15 +14,22 @@ interface IAttendButtonProps {
...
 const UnattendButton: FC<IAttendButtonProps> = ({ eventId, isOnWaitList, waitListNumber }) => {
   const dispatch = useDispatch();
-  const signOff = () => dispatch(removeAttendeeByEventId(eventId));
+  const [addToast] = useToast({ type: 'success', duration: 5000 });
+  const signOff = async () => {
+    try {
+      await dispatch(removeAttendeeByEventId(eventId)).then(unwrapResult);
+      addToast('Du har blitt meldt av arrangementet');
+    } catch (err) {
+      addToast(`Noe gikk galt under avmeldelse, ERROR: ${err.message}`, { type: 'error' });
```
FluidSense, line 23: Same as above, I suggest considering if it makes sense to stick to either `async-await` and `try-catch` or just `.then()`

src/events/components/Attendance/AttendButton.tsx
kharann: see other answer

src/events/components/Attendance/UnattendButton.tsx
kharann: I've decided that i think that its more readable with a try catch. I however removed the usage of then, instead i dispatch outside of the try-catch and unwrap it inside of the try-catch

FluidSense (approved):
Cool! Now I've learned about `unwrapResult` too!

PR #499, opened 2020-10-07
Remove text transform to fix captialized words which should not be capitalized

### LINKED ISSUE #428: Capitalized Letters In Event Status Text.
**Describe the bug**  
Status text "Du Må Være Logget Inn For Å Se Din Status." on events have capitalized first letters, that's not a Norwegian way to type it.

**To Reproduce**
1. Go to any event while not logged in
2. Look at the text. Look at it.

**Expected behavior**
Only the word "Du" to have a capitalized first letter.

tokongs:
# Changes
* Add style to the specific p tag which should not be capitalized.

Fixes #428

### DIFF
--- a/src/events/components/Attendance/index.tsx
+++ b/src/events/components/Attendance/index.tsx
@@ -16,7 +16,7 @@ const Attendance: FC<IAttendButtonProps> = (props: IAttendButtonProps) => {
   const currentTime = DateTime.local();
 
   // The canAttend-object is null if the user is not logged in, else it will always be present.
-  if (!canAttend) return <p style={{ textTransform: 'none' }}>Du må være logget inn for å se din status.</p>;
+  if (!canAttend) return <p>Du må være logget inn for å se din status.</p>;
 
   if ((currentTime < unattendDeadline && is_attendee) || is_on_waitlist) {
     return (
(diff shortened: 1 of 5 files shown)


kharann (dismissed):
There are quite a few other `<p>` tags which could need this same change too. I would maybe make a css-class and apply this className to all of the p tags.  (same file)

tokongs:
@kharann I don't think the other p tags that aren't supposed to be are affected by the capitalization.

oleast:
Instead of force-removing the style inline, should it not just be removed from the CSS?
https://github.com/dotkom/onlineweb-frontend/blob/c04dd2016ea85ba68e77d67a26d03fdbe0feb6a7/src/events/components/DetailView/detail.less#L69

tokongs:
@oleast I think it sort of makes sense to have it, as it captializes the data and stuff.

FluidSense:
I think there are multiple <p>-tags that are affected by this, I seem to recall reading "Du Er Nr 16 På Venteliste" which feels an awfully lot like 
4hr8vf

I think I support @oleast's suggestion of removing the rule, and rather create a class-rule or something which can be applied to the ones where we actually want it, to avoid confusion when creating new <p>-tags inside the DetailView

kharann:
> I think there are multiple `<p>`-tags that are affected by this, I seem to recall reading "Du Er Nr 16 På Venteliste" which feels an awfully lot like
> 4hr8vf
> 
> I think I support @oleast's suggestion of removing the rule, and rather create a class-rule or something which can be applied to the ones where we actually want it, to avoid confusion when creating new `<p>`-tags inside the`DetailView`

Takk Johannes

tokongs:
@FluidSense That sounds like a good plan, is the way it's done now the right way to do it?

FluidSense (approved):
Nice!

Issue #412, opened 2020-09-13
Handle Next-warnings


FluidSense:
Currently, we are prompted with 3 warnings on setup.

~~1. FrontPage returned an empty object from `getInitialProps`. This de-optimizes and prevents automatic static optimization. https://err.sh/vercel/next.js/empty-object-getInitialProps~~
2. Please do not use `@import` CSS syntax in createGlobalStyle at this time, as the CSSOM APIs we use in production do not handle it well. Instead, we recommend using a library such as react-helmet to inject a typical <link> meta tag to the stylesheet, or simply embedding it manually in your index.html <head> section for a simpler app.
3. Warning: viewport meta tags should not be used in _document.js's <Head>. https://err.sh/next.js/no-document-viewport-meta

FluidSense:
Number 2 could originate from[ '@dotkomonline/design-system'](https://github.com/dotkom/design-system/issues/184)

aslakhol:
2 is resolved by #502 and dotkom/design-system#190 

3. is resolved by #502

aslakhol:
1. is moved to #503

FluidSense:
Resolved in favor of #503

PR #502, opened 2020-10-09
Upgrade ds and include fonts

aslakhol:
Version 0.20.0 of DS was broken so we skipped it and went to 0.21.0 which needs us to include the fonts in this repo. 

In addition, I fixed the viewport in _document warning.


aslakhol:
resolves #412

FluidSense (approved):

PR #505, opened 2020-10-11
Fixed statistics donut-labels displaying ID instead of name

juliangra:
The donut was displaying event-type ID instead of the actual name of the event. This is now fixed.


FluidSense (approved):
LGTM!

PR #511, opened 2020-10-14
Avoid partial event attendees

### LINKED ISSUE #507: Event statistics shows partial event attendees which is not an option
**Describe the bug**  

Event statistics "commit"-view shows floats instead of integers.
This makes little sense as we do not have the option to attend half an event.

**To Reproduce**
1. Go to https://online.ntnu.no/profile/statistics/events
2. Scroll down to 'Arrangementsdatoer'

**Expected behavior**

Events being whole numbers

**Screenshots**

CarlOfHoly:
Sets the item count for each nivo-calender by finding the max possible value.

closes #507

### DIFF
--- a/src/common/components/Charts/CalendarChart.tsx
+++ b/src/common/components/Charts/CalendarChart.tsx
@@ -50,11 +50,7 @@ const CalendarChart = ({ frequency, header }: IProps) => {
   });
   const values: CalendarDatum[] = Object.keys(inter2).map((key) => ({ day: key, value: inter2[key] }));
 
-  const itemCount =
-    Math.max.apply(
-      Math,
-      values.map((calenderDatum) => calenderDatum.value)
-    ) - 1;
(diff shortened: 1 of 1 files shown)


src/common/components/Charts/CalendarChart.tsx
```diff
@@ -23,12 +23,12 @@ const MARGINS = {
...
 };
 
-const LEGENDS: CalendarLegend[] = [
+const LEGENDS = (itemCount: number): CalendarLegend[] => [
   {
     anchor: 'bottom-right',
     direction: 'row',
     translateY: 36,
-    itemCount: 4,
+    itemCount,
```
FluidSense, line 31: Wouldn't this have to be `itemCount: itemCount`?

FluidSense: ...apparently not 😂

FluidSense (approved):
Nice! I've noticed that the grid is gone (if we ever had one..!), but that could be another PR!

PR #512, opened 2020-10-15
Add /offline as overview-page for offline magazine

FluidSense:
We have an "archive" of old offlines at old, and would like for this to be present at OWF as well.
This is nice to have if you want to view old offlines without navigating the frontpage.

### DIFF
--- a/src/offline/index.tsx
+++ b/src/offline/index.tsx
@@ -1,0 +1,31 @@
+import React, { useEffect, useState } from 'react';
+
+import { getOfflines, getRemainingOfflines } from 'frontpage/api/offline';
+import { IOfflineIssue } from 'frontpage/models/Offline';
+import OfflineList from './components/OfflineList';
+import Spinner from 'common/components/Spinner';
+
+export const Offline: React.FC = () => {
+  const [offlines, setOfflines] = useState<IOfflineIssue[]>([]);
+
+  /** Get the first batch of Offlines for feast loading */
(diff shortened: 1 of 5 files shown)


src/offline/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { getOfflines, getRemainingOfflines } from 'frontpage/api/offline';
+import { IOfflineIssue } from 'frontpage/models/Offline';
+import OfflineList from './components/OfflineList';
+import Spinner from 'common/components/Spinner';
+
+export const Offline: React.FC = () => {
+  const [offlines, setOfflines] = useState<IOfflineIssue[]>([]);
+
+  /** Get the first batch of Offlines for feast loading */
```
oleast, line 11: Loading is a feast 🥳 🎈

FluidSense: Hahaha, I shamelessly copied this file from frontpage for faster development, but it should probably be refactored to share as much as possible with frontpage. Noticed the comment there, but forgot to fix it so now it is present in two files 😂

oleast: Yeah, I had i feeling I was the original sinner in this case 🙈

FluidSense:
Opened an issue (#513) to redo this view, so that prokom can refer to it in Offline and we can rather revisit this, the _agile_ way

aslakhol (approved):

Issue #462, opened 2020-09-24
Membership applications


aslakhol:
**Is your feature request related to a problem? Please describe.**
It's not possible to apply for memberships through OWF.

**Describe the solution you'd like**
It should be possible

**Describe alternatives you've considered**
Making https://old.online.ntnu.no/profile/membership/ but in OWF

aslakhol:
Extending memberships is relatively convoluted. Right now we need people to "lie" about their start date, and say that they started a year after they really did. This is both unintuitive and could make our users feel bad. 

We should handle this somehow. 

One possibility would be a rewrite of the system in OW. Perhaps looking at changing from start date to expected end date instead. 

Another possibility, that doesn't require big changes on the OW side, would be to make a button that does the lying for the user. The system would need to have information about the current membership and send a request that applies for the same membership, but with the shifted start date.

PR #517, opened 2020-10-16
Update, added membership panel in profile

juliangra:
Added a membership panel in profile settings with info regarding why you need membership, and a button linking to the membership application page on old.online.ntnu.no


src/profile/components/Settings/Membership/Info.tsx
```diff
@@ -0,0 +1,14 @@
+import Markdown from 'common/components/Markdown';
+import React from 'react';
+
+const ABOUT_PROFILE_MEMBERSHIP = `
+  # Medlemskap
+
+  Du trenger medlemskap for å kunne melde deg på linje-spesifikke arrangementer.
+
+  For å søke medlemskap, bruk knappen under
+`;
```
henrikhorluck, line 10: I don't see why this should be it's own component if the text is so tightly bound that it says "click the button below"

henrikhorluck (changes requested):

src/profile/components/Settings/Membership/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import MembersInfo from './Info';
+
+const Membership = () => {
+  return (
+    <Pane>
+      <MembersInfo></MembersInfo>
+      <Button
+        onClick={() => {
+          window.open('https://old.online.ntnu.no/profile/membership/');
+        }}
```
FluidSense, line 14: I think you could wrap the `Button` in a `Link`-component, as in `<Link href="https://old.online.ntnu.no/profile/membership/"><Button>Medlemskap</Button></Link>` component instead of an `onClick`-function! (You might need to add `passRef={true}` to the `Link`, not completely sure tho)
`window.open` has a [security flaw](https://stackoverflow.com/questions/40593632/use-window-open-but-block-use-of-window-opener) which the `Link`-component handles for us already!
You can import `Link` from `next/link`.

FluidSense (approved):
Nice work!

FluidSense:
Related to #462

PR #518, opened 2020-10-16
Refactor mail page 

kharann:
It currently looks like this

image

Any thoughts lads? I ain't not design god


FluidSense (approved):
🌾 very 📟 good 🚦 addition 🥌

PR #520, opened 2020-10-18
Remove incorrect README-instructions

FluidSense:
Also emphasized that the create rsa key-command is for OW4.


tokongs (approved):
LGTM :1st_place_medal:

PR #523, opened 2020-10-25
Fix attend button

### LINKED ISSUE #522: Change attend-button text to "Meld meg på venteliste" if the event is full
Maybe make it yellow too

Thomhas:
Fixed attend button when the event is full. 
Fixes #522


tokongs (approved):
LGTM!

Issue #524, opened 2020-10-25
Add infomail and opportunity sections to Mails


kharann:
This section is to get added to certain mailing lists, for opportunity or weekly summary

kharann:
Fixed in #525

PR #525, opened 2020-10-25
added infomail and opportunity section to mail

Thomhas:
Added infomail and opportunity section to mail + buttons to get added/removed to mailing lists. 
#524


kharann (approved):
LGTM

PR #526, opened 2020-10-25
Actually show opt-in useres as public on attendees list

### LINKED ISSUE #494: Shown as anonymous on events even when opt-in to show name on attendee list
**Describe the bug**  
Shown as anonymous on the attendee list even when I have checked in mine Settings-page that my name should på listed on events.

**To Reproduce**
1. Go to a event, sign up
2. Click on attendee list
3. See the error that almost everybody is anonymous, even if you have opt-in to be shown

**Expected behavior**
Name should be listed in attendee list if the checkbox is true

**Screenshots**

image
image

 - OS: Ubuntu
 - Browser Chrome

kharann:
Fixes #494

### DIFF
--- a/src/profile/models/Privacy.ts
+++ b/src/profile/models/Privacy.ts
@@ -4,7 +4,8 @@ export interface IPrivacy {
   expose_email: boolean;
   expose_phone_number: boolean;
   expose_address: boolean;
-  visible_as_attending_events: boolean;
+  visible_as_attending_events: boolean | null;
+  allow_pictures: boolean | null;
 }
 
 export const translations: { [Key in keyof IPrivacy]: string } = {
@@ -14,4 +15,5 @@ export const translations: { [Key in keyof IPrivacy]: string } = {
   expose_phone_number: 'Vis telefonnummer i brukerprofil',
   expose_address: 'Vis adresse i brukerprofil',
   visible_as_attending_events: 'Synlig på offentlige påmeldinglister',
+  allow_pictures: 'Tillate bilder at bilder av deg på offentlige arrangementer kan legges ut',
 };
--- a/src/profile/models/Privacy.ts
+++ b/src/profile/models/Privacy.ts
@@ -15,5 +15,5 @@ export const translations: { [Key in keyof IPrivacy]: string } = {
   expose_phone_number: 'Vis telefonnummer i brukerprofil',
   expose_address: 'Vis adresse i brukerprofil',
   visible_as_attending_events: 'Synlig på offentlige påmeldinglister',
-  allow_pictures: 'Tillate bilder at bilder av deg på offentlige arrangementer kan legges ut',
+  allow_pictures: 'Tillate at bilder av deg på offentlige arrangementer kan legges ut',
 };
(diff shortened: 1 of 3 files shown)


kharann:
I've  been thinking a bit now. Apperantly the API wants us to send in two privacy options, however i think that it might be better that it gets set in the backend from the user objects which we send in the call.

src/profile/models/Privacy.ts
```diff
@@ -14,4 +15,5 @@ export const translations: { [Key in keyof IPrivacy]: string } = {
   expose_phone_number: 'Vis telefonnummer i brukerprofil',
   expose_address: 'Vis adresse i brukerprofil',
   visible_as_attending_events: 'Synlig på offentlige påmeldinglister',
+  allow_pictures: 'Tillate bilder at bilder av deg på offentlige arrangementer kan legges ut',
```
oleast, line 18: "Tillate bilder at bilder"

oleast:
> I've been thinking a bit now. Apperantly the API wants us to send in two privacy options, however i think that it might be better that it gets set in the backend from the user objects which we send in the call.

Yeah, there is no problem with doing it that way, and we discussed it a bit when implementing the API.

The idea of requiring these options when signing up comes from the idea that the users should be presented with the choice **every time** they register for an event. Its not meant to be so that you have to just fetch the options and pass them through each time.
We wanted to have a checkbox in the registration form, like above or below the captcha, that would let you choose if you are visible _every time_. Then we could have your default from `Privacy` as the pre-filled value of the checkbox.

This also lets you do more interactive stuff as well. lets say you have the default in Privacy set to `False`, and you want to register for `Åre 2021`. You check the `Vis meg i påmeldingslisten` checkbox when you attend because you want your friends to be able to see that you are going, but haven't really checked out your profile before. Then we can have another checkbox or a button pop up and say 'Ønsker du å endre standardverdien for fremtidige arrangementer også?' to let users change the value then and there.

kharann:
> > I've been thinking a bit now. Apperantly the API wants us to send in two privacy options, however i think that it might be better that it gets set in the backend from the user objects which we send in the call.
> 
> Yeah, there is no problem with doing it that way, and we discussed it a bit when implementing the API.
> 
> The idea of requiring these options when signing up comes from the idea that the users should be presented with the choice **every time** they register for an event. Its not meant to be so that you have to just fetch the options and pass them through each time.
> We wanted to have a checkbox in the registration form, like above or below the captcha, that would let you choose if you are visible _every time_. Then we could have your default from `Privacy` as the pre-filled value of the checkbox.
> 
> This also lets you do more interactive stuff as well. lets say you have the default in Privacy set to `False`, and you want to register for `Åre 2021`. You check the `Vis meg i påmeldingslisten` checkbox when you attend because you want your friends to be able to see that you are going, but haven't really checked out your profile before. Then we can have another checkbox or a button pop up and say 'Ønsker du å endre standardverdien for fremtidige arrangementer også?' to let users change the value then and there.

Maybe we should merge in this temporarily and make privacy options in the modal a priority? Kinda of a bummer that users who have set their privacy options are not shown and the public list is just full of Anon users

oleast (approved):

PR #527, opened 2020-10-27
Rewrite ResponsiveImage to next/image for way smaller images

oleast:
So nextjs version 10 just dropped, and it includes some nice features.
One of these features is automatic image optimization, which we can use to make our images way smaller.

### DIFF
--- a/src/common/models/ResponsiveImage.ts
+++ b/src/common/models/ResponsiveImage.ts
@@ -5,6 +5,7 @@ export default interface IResponsiveImage extends IResponsiveImageSizes {
   description: string;
   tags: string[];
   photographer: string;
+  preset: ResponsiveImageTypes;
 }
 
 export interface IResponsiveImageSizes {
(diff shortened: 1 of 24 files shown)


src/common/models/ResponsiveImage.ts
```diff
@@ -17,7 +18,15 @@ export interface IResponsiveImageSizes {
...
-export type ResponsiveImageTypes = 'article' | 'company' | 'event' | 'offline' | 'product' | 'hobby' | 'resource';
+export type ResponsiveImageTypes =
+  | 'article'
+  | 'company'
+  | 'event'
+  | 'offline'
+  | 'product'
+  | 'hobby'
+  | 'resource'
+  | 'group';
```
henrikhorluck, line 29: Can't this be an enum? https://www.typescriptlang.org/docs/handbook/enums.html

s0lvang (approved):
lgtm

PR #542, opened 2021-02-28
adds ellipsis to end of paragraph to fix overflow

### LINKED ISSUE #166: Fix frontpage article cards when ingress is too long
A very long ingress can lead to problems with cards not displaying properly.

Example can be seen here, where the images on the smaller cards have whitespace below them.
frontpage-articles-2

Could be fixed by cutting a bit in the ingress text with overflow, as it already displays enough content in my opinion.

CarlOfHoly:
If the ingress is more than 10 lines, it stops the paragraph and adds an ellipsis (...)
closes #166


kharann (approved):
LGTM

Issue #553, opened 2021-04-22
Event view on mobile is too large on long words


FluidSense:
**Describe the bug**  

Event page is longer than mobile page if there are long links or long words.

The text block should have the CSS-rule: 
```css
word-break: anywhere;
hyphens: auto;
```

**To Reproduce**
1. Go to https://online.ntnu.no/events/1485
2. Open in responsive mode
3. The page is wider than the screen

**Expected behavior**

The text should be wrapped instead of being wider than the screen

**Screenshots**

With rule:
![image](https://user-images.githubusercontent.com/22146437/115718524-7ab63b00-a37b-11eb-977d-07714a2b626f.png)

Without rule (current):
![image](https://user-images.githubusercontent.com/22146437/115718569-8570d000-a37b-11eb-8ebd-2096205186f2.png)

FluidSense:
#554 Fixes for a lot of use cases, but according to MDN, the rules are not in use by Safari.

PR #555, opened 2021-04-24
Made word breaking only apply to links and improved browser support

HenrikSkog:
#553
Made css word breaking rules only apply to links and not all of the text to avoid unnatural word breaks. Also switched to rules that are better supported. 

Example of the unnatural word breaking that is currently happeninng on event pages:
image

Words of approximately this size "wwwwwwwwwwwwwwwwwwwwwww" that are not part of links still break the layout on mobile.

### DIFF
--- a/src/events/components/DetailView/detail.less
+++ b/src/events/components/DetailView/detail.less
@@ -88,8 +88,9 @@
   padding: 30px;
 }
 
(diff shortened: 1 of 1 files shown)


FluidSense (approved):
👍 ⚡

PR #563, opened 2021-09-12
✨ add Magnus to README

magnusrodseth:


README.md
```diff
@@ -70,3 +70,5 @@ You could also install the [prettier extension](https://marketplace.visualstudio
 ```
 
 This will enable prettier to auto-format on save for less and typescriptreact (.tsx files), and make sure that the stylelint linter is ran instead of the built-in validators
+
+Magnus
```
CarlOfHoly, line 74: Not good enough

CarlOfHoly (changes requested):

README.md
magnusrodseth: 😥

PR #571, opened 2021-09-30
Fixed attendees scrolling issue on Safari

juliangra:
- Added max-height to grid element to prevent overflow on Safari


kharann (approved):
Lgtm

PR #578, opened 2021-11-03
Update for company information text

AndersRobstad:
This change is just to make sure the correct name for our company events is used in the company information section. Since we have started using the name "bedriftsarrangementer" for both "kurs" and "bedriftspresentasjoner" it needs an update.


magnusrodseth (approved):
LGTM 🏎

PR #598, opened 2022-02-19
Link to new company page

henrikskog:
**NEWEST CHANGE**
- Remove "interessegrupper", "webshop" and "bidra" from header to make space for link
- This is obviously not uncontroversial
image

**CHANGELOG**
- Add new link in header that sends user to the new companies page located at new.online.ntnu.no/company
- Add Nextjs permanent redirect from /company to new.online.ntnu.no/company

**Not sure about**
- Might as well redirect directly to new.online.ntnu.no? Not sure if this is better, but I guess it does not matter that much.
- Link breaks the navbar design in desktop since there are too many links. Look at image. Could be fixed later just to get this page out fast. I am open to making changes if people have suggestions on fixes, eg. removing another link. 

image


henrikhorluck:
Suggestions: 
- Remove `Profil` from header, that is already available under the profile-icon/ "min side"
- Can "Ressurser" be removed, and it's individual components moved somewhere else? I think the following changes can be done to split the exisitng non-coherent version:
  - Move Discord to the footer
  - I think the info-pages don't need to be there? Would rather add it as part of the wiki?
  - SDF, "Studentressurser", Online-notifier (not touched since 2016 anyway) — remove
  - My largest concern is with "e-postlister" — perhaps it could be linken in the footer? Always found that a bit awkward to include with the rest of our services, as it has a different audience from most other parts of the webpage, perhaps also suitable for the wiki?

Issue #601, opened 2022-02-19
Empty address in profile page not formatted properly


HenrikSkog:
**Describe the bug**  
Empty address in profile page does not give "Ikke tilgjengelig", but it gives "null, null undefined"

**To Reproduce**
1. Go to https://online.ntnu.no/profile/settings and observe address field. You have to be signed in as a user which does not have their address defined.

**Expected behavior**
- Instead of undifined, null etc, there should be a default message if no address is provided, eg. "Ikke tilgjengelig", which is used on other fields. 

**Screenshots**
<img width="1107" alt="image" src="https://user-images.githubusercontent.com/55615149/154799430-4b4f2295-bc58-4f72-866c-995fa390f316.png">

PR #602, opened 2022-02-19
Fix/#601 profile empty address

HenrikSkog:
Closes: #601 

**Changelog**
1. Add util function that formats a string with variables, but that falls back to a default value if one of the variables are null/undefined.
2. Test function
3. Use function in Profile Settings where issue bug exists

**why**
- clean way to render the key value props in settings page in profile
- fixes issue where null/undifined are rendered to the website as strings

**example**
 * fmt('Navn: {0} {1}', 'Fornavn', 'Etternavn')) => 'Navn: Fornavn Etternavn'
 * fmt('Navn: {0} {1}', 'Fornavn', undefined)) => 'Ikke tilgjengelig'

**Possible issues**
- Bit of duplication of code, it is also checked for null values in the KeyValue component, but that does not work for strings with multiple variables
- **I have not been able to check that it works**
    - I have not been able to log in in development. 

**Important before merge**
- Check that it works in development


Thomhas (approved):
Looks good!

PR #615, opened 2022-03-08
fix: remove profile in header

HenrikSkog:
Profile link in header is redundant.

Changelog:
- Remove profile link from header


henrikhorluck (approved):
LGTM :shipit:

PR #631, opened 2022-08-13
dont redirect if URL is not landingpage

AndersRobstad:
Stops redirecting users when they visit online.ntnu.no/somepage, such that links to OW should work as intended. 

Not tested very much, but should work 🧑‍🚒


NjaalSoerland (approved):
LGTM

PR #632, opened 2022-08-13
Create committee application page

NjaalSoerland:
We need an application page running by tomorrow, and the server the current site is running on is down. I made a quick page to run temporarily until AMaCS is back up


NjaalSoerland:
LGTM

PR #647, opened 2022-11-26
Fix and improve extras select

Terbau:
# Changes

- Bumped `@dotkomonline/design-system` to `v0.22.2`
- Fixed events extras select component.
- Added a toast whenever a user selects a new extra.
- Extras (if exists) now automatically renders when you register for an event. Previously you had to refresh the page.

image
image

### DIFF
--- a/src/events/components/DetailView/AttendanceEvent.tsx
+++ b/src/events/components/DetailView/AttendanceEvent.tsx
@@ -55,13 +55,10 @@ const AttendanceEvent: FC<IProps> = ({ eventId, eventTitle }) => {
   const cancellationDeadline = DateTime.fromISO(attendanceEvent.unattend_deadline);
   const showPayment = !attendanceEvent.is_on_waitlist && attendanceEvent.is_attendee && attendanceEvent.payment;
 
-  const options: IExtraOption[] = [
-    { value: -1, label: 'Velg ekstra' },
-    ...attendanceEvent.extras.map((extra: IExtra) => ({
-      value: extra.id,
-      label: extra.choice,
-    })),
-  ];
+  const options: IExtraOption[] = attendanceEvent.extras.map((extra: IExtra) => ({
+    value: extra.id,
+    label: extra.choice,
+  }));
 
   return (
     <div className={style.blockGrid}>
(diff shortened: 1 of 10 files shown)


src/events/components/DetailView/AttendanceEvent.tsx
```diff
@@ -47,6 +55,14 @@ const AttendanceEvent: FC<IProps> = ({ eventId, eventTitle }) => {
   const cancellationDeadline = DateTime.fromISO(attendanceEvent.unattend_deadline);
   const showPayment = !attendanceEvent.is_on_waitlist && attendanceEvent.is_attendee && attendanceEvent.payment;
 
+  const options: IExtraOption[] = [
+    { value: -1, label: 'Velg ekstra' },
```
kharann, line 59: Shouldn't this just be a placeholder, instead of an option?

src/events/components/DetailView/AttendanceEvent.tsx
```diff
@@ -73,30 +89,15 @@ const AttendanceEvent: FC<IProps> = ({ eventId, eventTitle }) => {
...
-                onClick={(e: React.MouseEvent<HTMLOptionElement, MouseEvent>) =>
-                  dispatch(patchAttendee({ attendeeId: attendee.id, extras: Number(e.currentTarget.value) }))
-                }
-              >
-                {extra.id == attendee.extras && 'Valgt ekstra:'} {extra.choice}
-              </option>
-            ))}
-          </Select>
+        <Block title="Ekstras" className={style.fullBlock}>
+          <Select
```
kharann, line 93: as above, maybe use the placeholder prop

kharann (changes requested):

src/events/components/DetailView/AttendanceEvent.tsx
Terbau: I originally had the same thought, but I ended up using an option for it considering the backend api allows for resetting your choice of extras. If it was a placeholder, you would be locked into choosing an a valid option after one is already chosen, even if the event didn't necessarily require you to do so.

Terbau: But then again, it might be confusing and I'm not even sure if any events have ever had optional extras. I guess placeholder is the way to go, but I'd like another confirmation before I change it.

kharann: Don't think you should be able to reset your extras, even if it's possible in the API. I think that if an event would have an option from "None of the above", they should make it an explicit one.

Terbau: Yeah, that makes sense.

kharann (approved):

PR #692, opened 2023-09-18
fix performance issues events

henrikskog:
- only load participants when clicking show participants


henrikskog:
Anhkha godkjenner muntlig🧠

PR #694, opened 2023-09-27
replace mail addresses DOT-201

henrikskog:
Related: https://github.com/dotkom/onlineweb4/pull/3012


package.json
```diff
@@ -1,121 +0,0 @@
-{
```
junlarsen, line 1: Ooops? :sweat_smile:

junlarsen (changes requested):

junlarsen (approved):

PR #702, opened 2023-11-06
Add confirm modal on unattend event

Terbau:
Adds a confirm modal that appears whenever you click the unattend event button. This is to prevent users from accidentally unattending events.

## Preview
Desktop:
image

Mobile:
image


henrikskog (changes requested):
Very much needed, nice! Could we maybe call the function something like unregister or similar instead of signOff? 😇

henrikskog (approved):

PR #706, opened 2023-11-10
Disable renovate bot

jotjern:


junlarsen (changes requested):
We should just do like we did in https://github.com/dotkom/monoweb/commit/162f9b5a977ea38624765e16f3e97422e43d72da

PR #732, opened 2024-07-27
frontpage fixups

henrikskog:
- **set fixed width for event box on frontpage to avoid shifting**
- **push splash, dont set, so you can go back**
- **move styles into body to increase spesificity (stupid reset stuff)**
- **fix not centered images**


henrikskog:
[onlineweb-frontend-nrv27b06c-dotkom.vercel.app](https://onlineweb-frontend-nrv27b06c-dotkom.vercel.app/)

